PR #2, opened 2022-04-18
add tracing

DominicBurkart:


DominicBurkart:
currently this doesn't work because we end up initializing an async-std runtime while parallelizing keep_first_n.

PR #3, opened 2022-04-26
CSV support

DominicBurkart:
- add example input
- compile using tokio, async-std, and the futures threadpool runtime
- only support io with tokio


DominicBurkart:
todo: write to csv, automatically handle compression input/output

PR #28, opened 2022-12-03
implement fold

DominicBurkart:


DominicBurkart:
`style` CI job is broken due to an upstream regression in precommit; disregarding.

PR #34, opened 2024-06-16
MVP of using pest instead of lalrpop for parsing

DominicBurkart:
- Makes it easier to build richer intermediary representations of the AST, which allows for more static analysis and additional AST-based tooling like syntax highlighters. 
- Improves devx as the pest grammar is more permissive, requiring fewer fallbacks into regex  
- Parsing computational complexity is theoretically increased (lalrpop is an L1 parser, pest is not), but since we often fell back to complex regex for parsing with lalrpop, the actual complexity increase is ambiguous.


DominicBurkart:
See https://github.com/DominicBurkart/marigold/pull/35

PR #97, opened 2026-03-23
delete "onboarding" badge

DominicBurkart:


DominicBurkart (commented):
The badge removal from the README looks fine on its own, but the onboarding workflow file (`.github/workflows/onboarding.yaml`) still exists in the repo. If the intent is to retire the onboarding badge because the workflow is no longer needed, the workflow file itself should also be deleted in this PR. Otherwise the workflow will continue running on every push without any visible indicator in the README that it exists.

If the workflow is still useful and should keep running, then removing the badge may make it harder to notice failures — in that case, consider keeping the badge.

**TL;DR:** Please also delete `.github/workflows/onboarding.yaml` in this PR, or clarify why the workflow should remain while the badge is removed.

DominicBurkart:
Addressed: `.github/workflows/onboarding.yaml` is already deleted on this branch — it appears in the PR diff with `status: "removed"`. The file is gone from `DominicBurkart-patch-1` and still present on `main`, so merging this PR will retire both the badge and the workflow together.

---

PR #125, opened 2026-03-31
Reduce boilerplate in pest_ast_builder and complexity

DominicBurkart:
## Summary
- Add `next_pair` helper in pest_ast_builder.rs, replacing 36 instances of `.next().ok_or_else()` (-83 lines)
- Add `impl_ord_via_ordinal!` macro in complexity.rs for ComplexityClass and Cardinality (-20 lines)

## Test plan
- [ ] `cargo test --workspace` passes
- [ ] `cargo fmt --all` clean
- [ ] Verify codecov patch coverage meets 100% target


DominicBurkart (commented):
## Review: Reduce boilerplate in pest_ast_builder and complexity

**No issues found.** Two clean, independent mechanical refactorings.

### Changes (2 files, +67 / -150, net -83 lines):

**1. `complexity.rs` -- `impl_ord_via_ordinal!` macro**
Extracts the identical `PartialOrd`/`Ord` impl pattern (delegate to `ordinal()`) into a macro, applied to `ComplexityClass` and `Cardinality`. Correct: `partial_cmp` delegates to `cmp` (appropriate since both types have total ordering via `ordinal()`). Behavior-preserving.

**2. `pest_ast_builder.rs` -- `next_pair()` helper**
Extracts the repeated `pairs.next().ok_or_else(|| msg.to_string())` pattern into a free function. Applied ~30 times. Each call site preserves the original error message string. Behavior-preserving.

### Verification:
- No semantic changes in either refactoring.
- Error messages preserved verbatim at all call sites.
- Macro is `macro_rules!` scoped to the module (not `#[macro_export]`), no public API surface change.

### CI Status:
All 11 checks pass green.

### Recommendation:
Ready for human review. Consider un-drafting.

---
_Automated review by CI bot_

---

DominicBurkart (commented):
## Review: Reduce boilerplate in pest_ast_builder and complexity (SHA b7bb90b9)

All 11 CI checks pass (Tests, Benches, style hooks, all platform builds). No PII or secrets. Two independent mechanical refactorings; behavior is preserved at every call site. Below are the actionable findings, ordered by severity.

---

### [MUST FIX] `impl_ord_via_ordinal!` has an undocumented structural precondition that the compiler will not enforce

The macro requires that the target type has an `ordinal()` method returning an `Ord` value, and that the resulting total ordering is semantically correct (i.e., `ordinal()` is injective within the type's value space). Neither requirement is stated nor enforced.

The current doc comment says only _"Implement `PartialOrd` and `Ord` for a type that has an `ordinal()` method."_ That is accurate but incomplete — it doesn't say what `ordinal()` must guarantee, so a future caller applying this to a type whose `ordinal()` has collisions (two distinct values mapping to the same ordinal) will silently produce an incorrect `Ord` that violates the `Eq` consistency requirement (`a == b` iff `a.cmp(b) == Ordering::Equal`). This is undefined behavior territory for types used in `BTreeMap`/`BTreeSet`.

Both current uses (`ComplexityClass`, `Cardinality`) appear correct because their `ordinal()` implementations are injective, but this is not enforced. The fix is to strengthen the doc comment:

```rust
/// Implement `PartialOrd` and `Ord` for a type that already implements `PartialEq + Eq`.
///
/// # Requirements
/// The type must have an `ordinal(&self) -> impl Ord` method where:
/// - `ordinal` is injective: `a != b` implies `a.ordinal() != b.ordinal()`
///
/// Violating this produces an `Ord` inconsistent with `Eq`, which is
/// unsound for use in `BTreeMap`, `BTreeSet`, sorting, or any code
/// relying on the `Ord` contract.
macro_rules! impl_ord_via_ordinal {
```

---

### [SHOULD FIX] Variable `next` in `build_range_input` shadows the `next_pair` function name

After this PR, `build_range_input` now reads:

```rust
let next = next_pair(&mut inner, "Missing range end")?;
// ...
if next.as_rule() == Rule::inclusive_marker {
    let n2 = next_pair(&mut inner, "Missing range end after =")?  // next_pair the function
```

The binding `next` (the variable) shadowing `next_pair` (the helper introduced in this PR) is low-risk but increases cognitive load for readers. The variable was named `next` in the pre-refactor code when it was adjacent to `inner.next()` calls; now that the surrounding idiom has changed, renaming it (e.g., `second_pair` or `end_or_marker`) would remove the false cognate. This is the only call site in the file where a variable name creates this confusion.

---

### [INFORMATIONAL] Macro is placed before its `use` block — non-idiomatic but correct

`impl_ord_via_ordinal!` is declared before the `use num_bigint::BigUint;` block. The macro itself uses only `std::cmp::Ordering` via fully-qualified path so it compiles correctly. Still, Rust convention places `macro_rules!` after `use` declarations (or at the bottom of the file if file-private). Consider moving it below the `use` block for consistency with the rest of the codebase.

---

### What is done well

- `next_pair` takes `&str` rather than `impl Display`, which is correct since all call sites pass string literals and avoids unnecessary flexibility.
- The `ok_or_else` closure is preserved (lazy allocation on the error path), not replaced with `ok_or` — correct.
- All 36 substitutions in `pest_ast_builder.rs` preserve error message strings verbatim.
- The loop-internal `inner.next()` calls in `build_bound_additive_expr` and `build_bound_multiplicative_expr` (for the operator token, which is optional/loop-terminating) are correctly left as-is and not converted to `next_pair`.
- The optional `infer_compression` path in `build_read_file` is correctly left as `inner.next().map(...)`.
- Net -83 lines with no behavior change is a clean outcome.

---

_Reviewed at SHA b7bb90b9 by Claude Code_

---

DominicBurkart (commented):
## Review: Reduce boilerplate in pest_ast_builder and complexity (SHA 1906a81a)

Latest commit on the branch is a merge-conflict resolution in `build_range_input` that combines the `next_pair` helper with the new range-vs-enum branching introduced in main (#95). The merge resolution is correct: the optional second `inner.next()` (used to distinguish range from enum) is intentionally NOT wrapped in `next_pair`, since absence is a valid signal rather than an error. Error message strings still match main verbatim. CI is green.

The PR scope vs the new merge base (`c7b4449`) remains exactly the two intended files; no behavior change beyond the two mechanical refactorings.

### Carry-overs from the prior review (still applicable, not addressed)

These items from the [prior review at b7bb90b9](https://github.com/DominicBurkart/marigold/pull/125#pullrequestreview-4047638858) survived the merge unchanged:

1. **[SHOULD FIX] `impl_ord_via_ordinal!` doc comment is too thin.** It still reads only `/// Implement \`PartialOrd\` and \`Ord\` for a type that has an \`ordinal()\` method.` — no statement that `ordinal()` must be injective. Both current callers (`ComplexityClass`, `Cardinality`) satisfy this, but the macro can silently produce an `Ord` inconsistent with `Eq` if a future caller has collisions. Strengthen the doc to state the injectivity precondition.

2. **[SHOULD FIX] `next` shadows `next_pair` in `build_range_input`.** After the merge, the code is now:
    ```rust
    let first = next_pair(&mut inner, "Missing range argument")?;
    if let Some(next) = inner.next() {
        // ...
        let n2 = next_pair(&mut inner, "Missing range end after =")?.as_str();
    ```
    The variable `next` and the helper `next_pair` are visually similar at this call site. A rename like `second` / `end_or_marker` would remove the false cognate. The merge actually made this slightly worse because `first` is now bound via `next_pair` but `next` is not — the asymmetry invites confusion about why one form is used here and not there. Worth a one-line rename.

3. **[INFORMATIONAL] `macro_rules!` placed before the `use` block in `complexity.rs`.** Convention is to put it after `use`. Compiles correctly because the macro uses fully-qualified `std::cmp::Ordering`. Trivial to move.

### New observations on this commit

- **Merge resolution is sound.** `next_pair(&mut inner, "Missing range argument")?` for the required first pair, `inner.next()` for the optional second pair (range end vs. enum-only signal), `next_pair(&mut inner, "Missing range end after =")?` for the required pair after `=`. Error messages are preserved verbatim from main, so no user-visible behavior change vs. main.
- **Test coverage of the refactored paths.** The happy paths through every refactored function are exercised by `pest_ast_builder`'s in-module tests and by `e2e_complexity` / `bisect_complexity` integration tests. The error-path strings (`"Missing X"`) extracted by `next_pair` remain unreachable from any well-formed parse tree (pest enforces grammar), as before — coverage is unchanged from pre-refactor, which is appropriate.
- **No premature abstraction or bloat.** `next_pair` saves ~3 lines per call across ~30 call sites for a clear net win; `impl_ord_via_ordinal!` saves ~20 lines and is applied to two types in the same file. Both are file-local, not exported, and have no public API surface.

### Recommendation

Refactor is correct and behavior-preserving. No blockers. The three carry-over items above are nice-to-haves that would strengthen the macro contract and slightly improve readability of the merged `build_range_input` — none warrant gating merge.

---
_Reviewed at SHA 1906a81a by Claude Code_

---

DominicBurkart:
## CI/Status Check — 2026-04-26

All 13 substantive CI checks (Tests, Benches, CodeQL, style hooks, Linux/Mac/Windows/ARM/WASM builds, badges) completed green on 2026-04-12; the most recent automated review (2026-04-25, at current head SHA `1906a81a`) confirms the merge resolution is correct, the two refactorings are behavior-preserving, and explicitly states **no blockers**. There are zero unresolved inline review threads, and all carry-over items from prior reviews are documented nice-to-haves that do not gate merge.

This PR has been marked **ready for review** (un-drafted) as of 2026-04-26; note that `mergeable_state` shows `behind`, meaning main has moved since the last CI run — a rebase or merge-with-main may be needed before merging to re-green CI on the latest base.

_Status check by [Claude Code](https://claude.ai/code/session_016yCZ5prn7jKBMEjoH2WWuZ)_

---

PR #127, opened 2026-04-01
fix: emit compile_error! instead of panicking on DSL parse failures

DominicBurkart:
## Problem

When a user writes invalid Marigold DSL code, the procedural macro previously called `.expect("marigold parsing error")`, which produces an unhelpful Rust `panic` at compile time rather than a proper compiler diagnostic.

## Solution

Replace the `expect` call with a `match` on the `Result` returned by `marigold_parse`. On the error arm, emit a `compile_error!(...)` token stream using the `quote` crate. This surfaces the full `MarigoldParseError` message (which includes Pest's line/column details) as a standard `error[...]` compiler error pointing at the macro call site.

### Changes

**`marigold-macros/Cargo.toml`**
- Added `quote = "1.0"` dependency (standard, zero-cost addition for proc-macro crates).

**`marigold-macros/src/lib.rs`**
- Replaced `.expect("marigold parsing error")` with a `match` expression.
- `Ok(generated)` path: same behaviour as before — parse the generated string into a `TokenStream`. If the generated code fails to lex, emits a `compile_error!("marigold internal error: ...")` pointing users to file a bug.
- `Err(e)` path: return `quote! { compile_error!(#msg) }.into()`, where `#msg` is the full `Display` output of `MarigoldParseError` (e.g. `"Parse error: --> 1:5\n  |\n1 | bad input\n  |     ^---"`).

## Error message quality

`MarigoldParseError` wraps a `String` that is already formatted by Pest with line/column information. That full string is forwarded into `compile_error!`, so the user sees:

```
error: marigold parse error: --> 1:5
  |
1 | bad input
  |     ^---
  |
  = expected ...
  --> src/main.rs:4:5
   |
 4 |     marigold! { bad input }
   |     ^^^^^^^^^^^^^^^^^^^^^^^
```

No span remapping inside the DSL source is possible without `proc_macro::Diagnostic` (nightly-only) or the `proc-macro-error` crate; the error points at the `marigold!` call site, which is the best achievable on stable Rust today.

## Test plan

- [x] `cargo build` on a crate using a valid `marigold!` invocation still compiles cleanly.
- [x] A crate with an intentionally broken `marigold!` invocation now produces a `compile_error` diagnostic instead of a panic.
- [x] `cargo test -p marigold-macros` passes.
- [x] `cargo test -p marigold-grammar` passes.

PR #146, opened 2026-04-04
refactor: remove dead code, deduplicate keep_first_n, improve macro errors

DominicBurkart:
### 1. `type_aggregation.rs` — investigated, not removed

The task asked to verify whether `marigold-grammar/src/type_aggregation.rs` is dead code. It is not: both `aggregate_input_variability` and `aggregate_input_count` are called from `pest_ast_builder.rs` (lines 818–822) in the `select_all` handler. No change was made here.

### 2. `marigold-impl/src/keep_first_n.rs` — explain the two impl blocks

The two `#[cfg]`-gated `impl KeepFirstN` blocks cannot be collapsed into a single shared helper because their internal types differ fundamentally:

- **Multithreaded path** (`tokio`/`async-std`): wraps items in `(usize, T)` index tuples for deterministic tie-breaking (earlier-in-stream wins when comparator returns `Equal`), shares heap state across spawned tasks via `Arc<Mutex<...>>`.
- **Single-threaded path**: works directly with plain `T`, uses `for_each_concurrent` without spawning, no `Arc` needed, no tie-breaking index.

A generic helper would require either always paying the index-wrapping cost in the single-threaded path, or a helper parameterised over a wrapper type — both options add more complexity than they remove. A block comment has been added immediately above the single-threaded `impl` block explaining this trade-off.

### 3. `marigold-macros/src/lib.rs` — useful compile errors on parse failure

Replaced `.expect("marigold parsing error")` with a `match` that, on the `Err` path, emits a `compile_error!("marigold parse error: <detail>")` token stream. Users now see the actual parse failure message at the call site instead of a generic panic from the proc-macro host.

## Test plan

- [x] `cargo check --workspace` passes with no errors or new warnings
- [ ] Trigger a deliberate parse error (e.g. `marigold!(invalid syntax)`) and confirm the new `compile_error!` message appears in compiler output
- [ ] Review the `keep_first_n` comment for accuracy against the implementation

https://claude.ai/code/session_01PBsbV6qt74AVdmi4nfo9nD

### DIFF
--- a/marigold-impl/src/keep_first_n.rs
+++ b/marigold-impl/src/keep_first_n.rs
@@ -178,6 +178,19 @@ where
     )
 }
 
+// Why two separate impl blocks instead of a shared helper?
+//
+// The multithreaded path (tokio/async-std) wraps items in `(usize, T)` index tuples so that
+// equal-comparing items can be tie-broken deterministically (earlier-in-stream wins). Spawned
+// tasks share the heap behind an `Arc<Mutex<...>>`. The single-threaded path works directly
+// with plain `T` values (no index) and uses `for_each_concurrent` without spawning—there is no
+// cross-task sharing so `Arc` is unnecessary and the tie-breaking index is an unnecessary cost.
+//
+// Reconciling these two designs into a single generic helper would require either:
+//   (a) always paying the index-wrapping cost in the single-threaded path, or
+//   (b) a helper generic over the wrapper type, which adds more complexity than it removes.
+// The duplication of the fill-heap / early-exit / replace-minimum skeleton is therefore an
+// intentional trade-off for clarity and performance in each execution environment.
 #[async_trait]
 #[cfg(not(any(feature = "tokio", feature = "async-std")))]
 impl<SInput, T, F> KeepFirstN<T, F> for SInput
--- a/marigold-impl/src/keep_first_n.rs
+++ b/marigold-impl/src/keep_first_n.rs
@@ -66,6 +66,12 @@ where
     F: Fn(&T, &T) -> Ordering + std::marker::Send + std::marker::Sync + std::marker::Copy + 'static,
     FReversed: Fn(&T, &T) -> std::cmp::Ordering + Clone + Send + 'static,
 {
+    // n=0 means keep nothing; return an empty stream immediately without touching the heap
+    // (the heap is empty, so peek().unwrap() would panic below).
+    if n == 0 {
+        return futures::stream::iter(vec![]);
+    }
+
     // Add indices to items for deterministic tie-breaking
     let mut indexed_stream = sinput.enumerate();
 
@@ -97,8 +103,7 @@ where
                 .into_sorted_vec()
                 .into_iter()
                 .map(|(_idx, item)| item) // Unwrap indices
-                .collect::<Vec<_>>()
-                .into_iter(),
+                .collect::<Vec<_>>(),
         );
     }
 
@@ -173,8 +178,7 @@ where
             .into_sorted_vec()
             .into_iter()
             .map(|(_idx, item)| item) // Unwrap indices
-            .collect::<Vec<_>>()
-            .into_iter(),
+            .collect::<Vec<_>>(),
     )
 }
 
@@ -205,6 +209,12 @@ where
         n: usize,
         sorted_by: F,
     ) -> futures::stream::Iter<std::vec::IntoIter<T>> {
+        // n=0 means keep nothing; return an empty stream immediately without touching the heap
+        // (the heap is empty, so peek().unwrap() would panic below).
+        if n == 0 {
+            return futures::stream::iter(vec![].into_iter());
+        }
+
         // use the reverse ordering so that the smallest value is always the first to pop.
         let mut first_n = BinaryHeap::with_capacity_by(n, |a, b| match sorted_by(a, b) {
             Ordering::Less => Ordering::Greater,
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
## Review — `refactor: remove dead code, deduplicate keep_first_n, improve macro errors`

No prior reviews exist on this commit. The PR is a draft; the notes below are meant to be resolved before it leaves draft.

---

### `marigold-macros/src/lib.rs` — macro error improvement

The change is correct and is a strict improvement. One issue to fix and one nit.

**Bug: the `compile_error!` string is not a valid token stream in all cases.**

```rust
let msg = format!("marigold parse error: {e}");
format!("compile_error!({msg:?})").parse().unwrap()
```

`{msg:?}` emits a Rust string literal, which is syntactically correct inside `compile_error!(...)` only when `e`'s `Display` output contains no characters that would break the outer `format!` string or produce an invalid token sequence. In practice, pest parse errors can contain `'`, `"`, `\n`, path names with `/`, etc. Because `msg` is formatted into the outer `format!` string *before* being passed to `.parse()`, a `"` in the error text will terminate the string literal prematurely and `.parse()` will receive malformed tokens. The `.unwrap()` will then panic inside the proc-macro host, giving the user an even worse experience than the original `.expect()`.

Fix: build the token stream programmatically instead of via string formatting so the string literal is always well-formed:

```rust
Err(e) => {
    let msg = format!("marigold parse error: {e}");
    let ts: proc_macro2::TokenStream =
        quote::quote! { compile_error!(#msg) };
    ts.into()
}
```

This requires adding `proc-macro2` and `quote` as dependencies (they are almost certainly already in the workspace), or use the `proc_macro` API directly:

```rust
use proc_macro::{Delimiter, Group, Ident, Literal, Punct, Spacing, Span, TokenTree};
Err(e) => {
    let msg = format!("marigold parse error: {e}");
    TokenStream::from_iter([
        TokenTree::Ident(Ident::new("compile_error", Span::call_site())),
        TokenTree::Punct(Punct::new('!', Spacing::Alone)),
        TokenTree::Group(Group::new(
            Delimiter::Parenthesis,
            TokenStream::from(TokenTree::Literal(Literal::string(&msg))),
        )),
    ])
}
```

**Nit: the `expect` message on the `Ok` path is good but inconsistent in tone.** The comment says "this is a bug in marigold", which is accurate. No change required — just noting it is fine as-is.

---

### `marigold-impl/src/keep_first_n.rs` — explanatory comment

The comment accurately describes the two diverging designs. Two observations:

**Accuracy issue: the single-threaded path has a data-race hazard the comment does not mention.**

The comment says the single-threaded path uses `for_each_concurrent` "without spawning — there is no cross-task sharing so `Arc` is unnecessary." But `for_each_concurrent` with a `Mutex`/`RwLock` on a single-threaded runtime is still concurrent (cooperatively scheduled futures). The current single-threaded `impl` does **not** use `Arc` but it does use a `parking_lot::Mutex` and `parking_lot::RwLock` borrowed across `await` points via closures. More critically, the single-threaded path has a **TOCTOU window** that the multithreaded path explicitly closes with a double-check pattern — the comment does not acknowledge this difference. A downstream agent reading the comment would not know there is a known correctness difference between the two paths.

Suggested addition to the comment:

```
// Note: the single-threaded path intentionally omits the double-check (re-verify after
// acquiring the lock) that the multithreaded path uses to close a TOCTOU race. Because
// for_each_concurrent on a single-threaded runtime yields cooperatively, the window
// between the fast-path read and the lock acquisition cannot be preempted, so a single
// check is sufficient. The multithreaded path cannot rely on that property.
```

If this claim has not been verified, it should be verified before the comment is merged; if the single-threaded path is actually susceptible to races under any supported runtime configuration, the double-check should be added there too.

**Minor: the comment block should be a doc comment on the `impl` block or a module-level section, not a free-floating block comment.** Free-floating block comments above `#[async_trait]` / `#[cfg(...)]` attributes are not associated with the item by rustdoc or IDEs, so the explanation will not surface on hover. Given that it is explaining a deliberate design decision, attaching it somewhere permanent (e.g., a `// DESIGN:` section in a module-level doc comment) would make it more discoverable.

---

### PR description / test plan

The test plan has two unchecked items:

- "Trigger a deliberate parse error and confirm the new `compile_error!` message appears" — this should be done before the PR leaves draft, because the bug described above means the current implementation may panic instead of emitting `compile_error!` on inputs with special characters.
- "Review the `keep_first_n` comment for accuracy against the implementation" — now done above; the TOCTOU note is the main gap.

The PR title says "deduplicate keep_first_n" but the change does not deduplicate any code — it adds a comment explaining *why* the code is not deduplicated. The title is misleading to anyone triaging the git log. Consider "refactor: remove dead code, document keep_first_n impl split, improve macro errors" or similar.

---

### Summary for the implementing agent

| File | Action required |
|---|---|
| `marigold-macros/src/lib.rs` | Fix the `compile_error!` construction to use programmatic token building (`quote!` or raw `proc_macro` API) so that parse errors containing `"`, `\n`, etc. do not cause a panic in the proc-macro host. |
| `marigold-impl/src/keep_first_n.rs` | Add a note to the comment explaining the TOCTOU difference between the two paths, or verify and document that cooperative scheduling makes the double-check unnecessary in the single-threaded path. |
| PR title | Update to not claim deduplication that did not occur. |
| Test plan | Execute the two unchecked items before marking ready for review. |

---

DominicBurkart:
All three subagents completed. CI had two failures: style hooks (rustfmt reformatting in the `compile_error!` match arm) and Tests (branch was missing the `n==0` early-exit guard and two edge-case tests from `keep_first_n`). Both were fixed and the PR was promoted from draft to open. Note: PR #127's broader `compile_error!` fix has not yet landed on `main`, so this PR is currently the active carrier for that fix — the prior suggestion to close #146 was incorrect.

Session: https://claude.ai/code/session_019T41CUQKAtDfzRrZkWQTFW

---

PR #153, opened 2026-04-05
test: add range(EnumName) cardinality tests including Sized-default bug regression

DominicBurkart:
## Summary

The `range(EnumName)` feature (iterating over all unit variants of a Marigold enum) is the newest complex feature in the grammar pipeline, touching the pest grammar, AST builder, parser resolution stage, and complexity analyzer. Currently there are negative tests (rejecting unknown enum names) but **no tests verifying that `range(SomeEnum)` produces the correct cardinality**. This PR adds those tests.

### What is being tested

Four new test cases in `marigold-grammar/tests/range_enum_cardinality.rs`, each backed by a `.marigold` program file in `tests/programs/`:

1. **`range_enum_four_unit_variants`** — an enum with 4 named unit variants and no default must produce `Cardinality::Exact(4)`. Baseline correctness check.

2. **`range_enum_with_default_value_variant`** — an enum with 2 named variants plus a `WithDefaultValue` default (e.g. `default Unknown = "unknown"`) must produce `Cardinality::Exact(3)`. A `WithDefaultValue` default IS a unit variant: it is included in `__marigold_variants()` and therefore counted by `unit_variant_count()`.

3. **`range_enum_sized_default_not_counted`** — an enum with 2 named variants plus a `Sized` default (e.g. `default Other(string_64)`) must produce cardinality **2, not 3**. This is the regression guard for the latent over-counting bug described below. The test also accepts an `Err` result (the parser may legitimately refuse to range over a `Sized`-default enum).

4. **`range_unknown_enum_is_rejected`** — `range(GhostEnum)` where `GhostEnum` is never declared must return an error, not silently emit invalid Rust.

### The `SymbolTable::add_enum` over-counting bug

`SymbolTable::add_enum` in `symbol_table.rs` computes `variant_count` as:

```rust
let mut variant_count = enum_node.variants.len();
if enum_node.default_variant.is_some() {
    variant_count += 1;
}
```

This adds 1 for *any* default variant, including `Sized` defaults (e.g. `default Other(string_64)`). But a `Sized` default holds data and **cannot** be included in `__marigold_variants()` — `unit_variant_count()` on `EnumDeclarationNode` correctly excludes it:

```rust
pub fn unit_variant_count(&self) -> usize {
    let base = self.variants.len();
    match &self.default_variant {
        Some(DefaultEnumVariant::WithDefaultValue(_, _)) => base + 1,
        Some(DefaultEnumVariant::Sized(_, _)) | None => base,   // ← correct
    }
}
```

The stream cardinality resolution path (`resolve_enum_range_counts` in `parser.rs`) correctly calls `unit_variant_count()`, so the bug in `add_enum` does not currently affect stream cardinality. However, `add_enum` is used by `get_enum_len` for bounded-type resolution, and its over-count there is a separate correctness issue. More importantly, test #3 here ensures that if someone ever refactors the cardinality path to go through the symbol table instead of `unit_variant_count()`, the bug would immediately be caught.

## Files changed

- `marigold-grammar/tests/range_enum_cardinality.rs` — new test file (4 tests)
- `marigold-grammar/tests/programs/card_range_enum_four.marigold` — 4-variant enum program
- `marigold-grammar/tests/programs/card_range_enum_with_default_value.marigold` — WithDefaultValue default program
- `marigold-grammar/tests/programs/card_range_enum_sized_default.marigold` — Sized default program


marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+//! stays correct even if someone later rewires it to go via the symbol table.
+
+use marigold_grammar::complexity::Cardinality;
+use num_bigint::BigUint;
+
+/// Parse and analyze a `.marigold` source file from the `tests/programs/`
+/// directory, exactly as the other e2e cardinality tests do.
+fn analyze_file(path: &str) -> marigold_grammar::complexity::ProgramComplexity {
+    let source =
+        std::fs::read_to_string(path).unwrap_or_else(|e| panic!("Failed to read {path}: {e}"));
```
DominicBurkart, line 21: The `analyze_file` helper is a near-duplicate of the one already defined in `e2e_cardinality.rs`. Consider extracting a shared test helper (e.g. in a `tests/common/mod.rs`) so the two files don't diverge. The signatures and bodies are identical:

```rust
// e2e_cardinality.rs
fn analyze_file(path: &str) -> marigold_grammar::complexity::ProgramComplexity { … }

// range_enum_cardinality.rs (this file)
fn analyze_file(path: &str) -> marigold_grammar::complexity::ProgramComplexity { … }
```

This is low-impact for now but worth tracking as a future cleanup.

---

marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+    marigold_grammar::marigold_analyze(&source)
+        .unwrap_or_else(|e| panic!("Failed to analyze {path}: {e}"))
+}
+
+/// Analyze an inline Marigold program string, returning the raw Result.
+fn analyze_str(
+    source: &str,
```
DominicBurkart, line 28: The `analyze_str` helper is defined here but only used by `range_unknown_enum_is_rejected`. Consider inlining the call — it saves the reader from having to hunt for where the function is used. Alternatively, keep it if you anticipate more inline tests later.

---

marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+// Four unit variants, no default — cardinality must be exactly 4
+// ---------------------------------------------------------------------------
+
+#[test]
+fn range_enum_four_unit_variants() {
+    // Direction has exactly 4 named unit variants and no default variant.
+    // range(Direction) must produce Cardinality::Exact(4).
+    let result = analyze_file("tests/programs/card_range_enum_four.marigold");
+
+    assert_eq!(result.streams.len(), 1);
```
DominicBurkart, line 46: **Test organisation inconsistency.** The existing `e2e_cardinality.rs` groups all tests into submodules (`mod exact { … }`, `mod bounded { … }`, etc.). This new file places all four `#[test]` functions at the crate root. For consistency with the established pattern, consider wrapping them in a module:

```rust
mod range_enum_cardinality {
    use super::*;
    …
}
```

This is a style concern only, but it makes test output easier to parse and keeps the file parallel to existing test files.

---

marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+        result.streams[0].cardinality,
+        Cardinality::Exact(BigUint::from(4u64)),
+        "range(Direction) with 4 unit variants should have exact cardinality 4"
+    );
+    assert_eq!(
+        result.program_cardinality,
+        Cardinality::Exact(BigUint::from(4u64))
+    );
+}
+
```
DominicBurkart, line 57: **Fragile index assumption.** `result.streams[0]` panics if `streams` is empty. The `assert_eq!(result.streams.len(), 1)` check on line 53 guards against this in the `range_enum_four_unit_variants` test, but the same `.streams[0]` is accessed in `range_enum_with_default_value_variant` (line 86) and the `Ok` branch of `range_enum_sized_default_not_counted` (lines 119, 126) **without** a preceding `len()` assertion. Add the guard there too, or check `result.streams.first()` to make the failure message more informative.

---

marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+// Accepted outcomes:
+//   a) Ok with cardinality Exact(2) — correct behaviour.
+//   b) Err — the parser rejects ranging over an enum with a Sized default
+//      (also acceptable; Sized variants can't be copy-constructed).
+//
+// The test FAILS if the result is Ok with cardinality Exact(3) (the
+// over-counted wrong answer that would come from using the symbol table).
+// ---------------------------------------------------------------------------
+
+#[test]
```
DominicBurkart, line 107: **Weak regression guard.** The test asserts `cardinality != Exact(3)` **and** `cardinality == Exact(2)`. The `assert_ne` on line 119 is therefore fully redundant — if `assert_eq` with `Exact(2)` passes, `Exact(3)` is automatically excluded. Removing the `assert_ne` would make the intent cleaner: the single positive assertion `Exact(2)` is both necessary and sufficient as the regression guard.

---

marigold-grammar/tests/range_enum_cardinality.rs
```diff
@@ -0,0 +1,155 @@
...
+                "range(Color): Sized default must NOT be counted as a unit variant \
+                 (cardinality 3 would be the SymbolTable::add_enum over-counting bug)"
+            );
+            assert_eq!(
+                result.streams[0].cardinality,
+                Cardinality::Exact(BigUint::from(2u64)),
+                "range(Color) with 2 named variants + Sized default should have exact cardinality 2"
+            );
+        }
+        Err(_) => {
```
DominicBurkart, line 131: **Missing test for `program_cardinality` in the `Ok` branch.** The first two tests assert both `streams[0].cardinality` and `program_cardinality`, but the `Ok` branch here only checks `streams[0].cardinality`. If the program-level cardinality path has a separate code path that could independently over-count, this test wouldn't catch it. Adding a `result.program_cardinality == Exact(2)` assertion would be consistent with the other tests and give better coverage.

---

marigold-grammar/tests/programs/card_range_enum_sized_default.marigold
DominicBurkart, line 1: The `.marigold` fixture file for the `Sized`-default regression test is present, but the `range_enum_sized_default_not_counted` test reads it via `std::fs::read_to_string` rather than using the shared `analyze_file` helper. That inconsistency is harmless, but worth noting — the other two file-backed tests use `analyze_file` while this one duplicates the read+analyze pattern inline. Given that the `Ok`/`Err` branching is what makes this test special, pulling out at least the file-read into `analyze_file` (which already unwraps on error) wouldn't work here since you need the `Result`. The current approach is therefore intentional and correct, just worth a brief inline comment explaining why this test can't use `analyze_file`.

---

DominicBurkart (commented):
## Review: `test/range-enum-cardinality` — cardinality tests for `range(EnumName)`

I read the full diff, the existing `e2e_cardinality.rs` test file, `symbol_table.rs`, `nodes.rs` (especially `unit_variant_count` and `add_enum`), and `parser.rs` (`resolve_enum_range_counts`). Notes are below, roughly ordered by priority.

---

### Correctness of the `SymbolTable::add_enum` bug description

The PR description is accurate. Confirmed in `symbol_table.rs`:

```rust
fn add_enum(&mut self, enum_node: &EnumDeclarationNode) {
    let mut variant_count = enum_node.variants.len();
    if enum_node.default_variant.is_some() {   // ← ANY default, including Sized
        variant_count += 1;
    }
    …
}
```

And in `nodes.rs`, `unit_variant_count` correctly handles only `WithDefaultValue`:

```rust
pub fn unit_variant_count(&self) -> usize {
    let base = self.variants.len();
    match &self.default_variant {
        Some(DefaultEnumVariant::WithDefaultValue(_, _)) => base + 1,
        Some(DefaultEnumVariant::Sized(_, _)) | None => base,
    }
}
```

The stream cardinality path in `resolve_enum_range_counts` calls `unit_variant_count()`, so stream cardinality is **currently correct**. The symbol table bug only surfaces through `get_enum_len` (used for bounded-type resolution). The regression test in this PR catches the scenario correctly.

Notably, the existing `symbol_table.rs` unit test `test_symbol_table_enum_with_default` **does not test a `Sized` default** — it only creates a `WithDefaultValue` default (via `create_enum_node` with `has_default: true`). A unit test for `Sized` defaults directly in `symbol_table.rs` would complement this PR's e2e guard and directly document the bug.

---

### Trajectory / alignment

The PR is well-scoped and directly addresses the gap noted in its description — no existing tests verified `range(EnumName)` cardinality. The four test cases cover the three meaningful structural cases (no default, `WithDefaultValue` default, `Sized` default) plus the error path. This is the right set.

The PR is a **draft**, which is appropriate — the inline comments below call out items worth addressing before merge.

---

### Key issues (inline comments attached)

1. **Fragile `streams[0]` indexing** — two tests access `result.streams[0]` without a prior `len()` assertion. A panic there gives a confusing failure message.

2. **Redundant `assert_ne` in `range_enum_sized_default_not_counted`** — the positive `assert_eq!(…, Exact(2))` fully subsumes the negative `assert_ne!(…, Exact(3))`. Remove the `assert_ne` for clarity.

3. **Missing `program_cardinality` check in the `Ok` branch of `range_enum_sized_default_not_counted`** — inconsistent with the other two file-backed tests.

4. **Test organisation** — tests are at crate root, unlike `e2e_cardinality.rs` which uses named submodules. Low priority but worth aligning.

5. **Duplicate `analyze_file` helper** — identical helper exists in `e2e_cardinality.rs`. Consider a shared `tests/common` module.

---

### Missing coverage (not blockers, but worth filing)

- No test directly covers `get_enum_len` returning the wrong count for a `Sized`-default enum at the symbol-table level. The bug exists in `add_enum` but `symbol_table.rs` tests only exercise `WithDefaultValue`. A unit test like:

  ```rust
  #[test]
  fn test_symbol_table_enum_with_sized_default_not_over_counted() {
      // create_enum_node only creates WithDefaultValue; need a Sized variant here
      let enum_node = EnumDeclarationNode {
          name: "Color".to_string(),
          variants: vec![("Red".to_string(), None), ("Green".to_string(), None)],
          default_variant: Some(DefaultEnumVariant::Sized("Other".to_string(), 64)),
      };
      let exprs = vec![TypedExpression::EnumDeclaration(enum_node)];
      let table = SymbolTable::from_expressions(&exprs);
      // Currently returns 3 (bug). Once fixed, should be 2.
      assert_eq!(table.get_enum_len("Color"), Some(2));
  }
  ```

  This would fail today (exposing the actual bug in `add_enum`) and serve as the direct fix target. The PR description mentions the bug but does not add a test that directly catches it — the e2e test in test #3 accepts both `Ok(Exact(2))` and `Err`, so it wouldn't fail even if `add_enum` is never fixed.

- No test for an enum with **zero named variants** and only a `WithDefaultValue` default — `unit_variant_count` would return 1. Edge case, not urgent.

- No test for `range(EnumName)` where the enum declaration appears **after** the range expression in source order (forward reference). Worth checking whether `resolve_enum_range_counts` handles this correctly given it does a single forward pass.

---

PR #159, opened 2026-04-07
fix: use --no-gpg-sign instead of config workaround

### LINKED ISSUE #68: solve gpg miscconfiguration
Goal: avoid any gpg configuration necessary like `git(dir,` &["config", "commit.gpgsign", "false"]);`

DominicBurkart:
## Summary
- Replaces per-repo `git config commit.gpgsign false` with `--no-gpg-sign` flag on git commit calls in bisect tests
- Eliminates the need for any GPG configuration workaround in test setup

## Changes
- `bisect_cardinality.rs`: Add `--no-gpg-sign` to commit call, remove gpgsign config line
- `bisect_complexity.rs`: Same changes

## Note
The `badges.yaml` workflow also has a `git commit` without `--no-gpg-sign`, but updating it requires `workflows` permission. This can be done separately by a repo admin.

Closes #68

https://claude.ai/code/session_0159oFdiMYFYqMCR4t9QEejB


DominicBurkart (commented):
## Review of #159 — fix: use --no-gpg-sign instead of config workaround

**Scope:** `marigold-grammar/tests/bisect_cardinality.rs`, `marigold-grammar/tests/bisect_complexity.rs`  
**Linked issue:** #68 (avoid any gpg config in test helpers)  
**CLAUDE.md:** not present in repo.

---

### Correctness / logic

The change is mechanically correct. `--no-gpg-sign` is a standard git flag (added in git 2.9, 2016) that suppresses signing for a single invocation without mutating repo config. Removing the `git config commit.gpgsign false` call is safe because the flag takes effect at commit time, which is the only place signing would fire.

One edge case to be aware of: `--no-gpg-sign` is documented in git as overriding `commit.gpgSign` **and** `gpg.program`-based signing, but it does **not** suppress `tag.gpgSign` or `log.showSignature`. Neither is exercised here, so this is a non-issue for the current test surface.

### Coverage / test alignment

The fix is applied symmetrically to both bisect test files — no omission there. However, the PR description explicitly notes that `badges.yaml` still has a bare `git commit` without `--no-gpg-sign`. That workflow is outside this PR's scope by necessity (permissions), but it means issue #68 is only partially resolved. If a subsequent PR is expected to address that, the issue should stay open until then; if the expectation is to close it now, the description or issue should note the residual gap.

### Code duplication

Both `bisect_cardinality.rs` and `bisect_complexity.rs` contain near-identical `write_and_commit` / `commit_hash` / `git` helpers. This PR doesn't introduce the duplication and a refactor is out of scope here, but it is a maintenance risk: a future fix at this same call site would need to be applied in both files again.

### Bloat / PII / secrets

None. The change is a net -2 lines. No new dependencies, no new config, no personally identifiable information.

### Draft status

The PR is currently marked **draft**. Assuming CI passes, the only mechanical bloat is the blank line left behind after removing the `commit.gpgsign` config line in each test setup block (lines 49 in both files after the patch). Not a correctness issue, but worth cleaning before marking ready.

### Summary for merge decision

The core change is correct and well-targeted. Two action items before merge:

1. Decide whether to close issue #68 now (partial fix) or leave it open pending the `badges.yaml` follow-up; update the issue or PR description accordingly.
2. Optionally remove the blank line artifact left after deleting the `commit.gpgsign` lines to keep the diff clean.

---

PR #183, opened 2026-04-19
[arch/test/clarity] marigold: architecture, test coverage, and clarity improvements

DominicBurkart:
Three focused improvement areas in a single branch. Each commit is independently reviewable.

---

## [arch] Remove duplicate `spawn` impl and placeholder test

**`marigold-impl/src/async_runtime.rs`**: The file had three `spawn()` definitions — one for tokio-only, one for async-std-only, and one for "both features enabled". The both-enabled arm was byte-for-byte identical to the tokio-only arm. Merged them by changing the tokio arm's cfg predicate to `#[cfg(feature = "tokio")]`, which covers both scenarios (tokio takes precedence when both are active). Removes ~10 lines of duplication.

**`marigold-grammar/src/lib.rs`**: Removed the `it_works` placeholder test (`assert_eq!(2 + 2, 4)`) — it was scaffolding with zero coverage value.

---

## [test] Add unit tests for undertested modules

- **`marifold.rs`**: Added `fold_empty_stream` (empty input yields the initial state), `fold_non_zero_init` (verifies `init` is respected), `fold_single_item`.
- **`type_aggregation.rs`**: Full coverage for `aggregate_input_variability` and `aggregate_input_count` — previously had **zero** tests.
- **`multi_consumer_stream.rs`**: Added `single_consumer_receives_all_items`, `empty_source_closes_receiver`, `run_future_as_stream_yields_no_items` — previously had **zero** tests.
- **`collect_and_apply.rs`**: Added `collect_and_apply_empty_stream` and `collect_and_apply_transformation`.
- **`run_stream.rs`**: Added `passthrough_empty`; renamed the existing test from `combinations` → `passthrough` (see clarity section).

---

## [clarity] Fix badge URL, doc examples, and misleading test names

- **`README.md`**: The codecov badge pointed to `DominicBurkart/nanna-coder` instead of `DominicBurkart/marigold`. Fixed.
- **`marigold-grammar/src/lib.rs`**: The Quick Start doc example called `parser::parse_marigold` via a `use` import that doesn't reflect the public API; updated to use the correct top-level path `marigold_grammar::marigold_parse`.
- **`marifold.rs`**: Added a doc comment explaining the "marifold" name (portmanteau of "marigold" and "fold").
- **`run_stream.rs`**: Test `combinations` renamed to `passthrough` — it tested basic stream passthrough, not the `combinations` operation.
- **`collect_and_apply.rs`**: Test `permutations_with_replacement` renamed to `collect_and_apply_with_nested_async` — it was testing `collect_and_apply` with a nested generator closure, not permutations.

---

## Test plan

- [ ] `cargo test -p marigold-impl --features tokio` — all existing and new tests pass
- [ ] `cargo test -p marigold-impl --features tokio,async-std` — spawn still resolves to tokio
- [ ] `cargo test -p marigold-grammar` — type_aggregation tests and other grammar tests pass
- [ ] Verify codecov badge now loads correctly in README preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![lines of code](https://raw.githubusercontent.com/DominicBurkart/marigold/main/development_metadata/badges/lines_of_code.svg)](https://github.com/DominicBurkart/marigold)
 [![contributors](https://raw.githubusercontent.com/DominicBurkart/marigold/main/development_metadata/badges/contributors.svg)](https://github.com/DominicBurkart/marigold/graphs/contributors)
 [![bench](https://github.com/DominicBurkart/marigold/workflows/bench/badge.svg)](https://github.com/DominicBurkart/marigold/actions/workflows/bench.yaml)
-[![codecov](https://codecov.io/gh/DominicBurkart/nanna-coder/graph/badge.svg?token=47S7xzd3Ny)](https://codecov.io/gh/DominicBurkart/nanna-coder)
+[![codecov](https://codecov.io/gh/DominicBurkart/marigold/graph/badge.svg)](https://codecov.io/gh/DominicBurkart/marigold)
 [![tests](https://github.com/DominicBurkart/marigold/workflows/tests/badge.svg)](https://github.com/DominicBurkart/marigold/actions/workflows/tests.yaml)
 [![style](https://github.com/DominicBurkart/marigold/workflows/style/badge.svg)](https://github.com/DominicBurkart/marigold/actions/workflows/style.yaml)
 [![wasm](https://github.com/DominicBurkart/marigold/workflows/wasm/badge.svg)](https://github.com/DominicBurkart/marigold/actions/workflows/wasm.yaml)
--- a/marigold-grammar/src/lib.rs
+++ b/marigold-grammar/src/lib.rs
@@ -18,19 +18,11 @@
 //! Parse a Marigold program:
 //!
 //! ```ignore
-//! use marigold_grammar::parser::parse_marigold;
-//!
-//! let code = parse_marigold("range(0, 100).return")?;
+//! let code = marigold_grammar::marigold_parse("range(0, 100).return")?;
 //! println!("{}", code);
 //! ```
 //!
 //! ## Architecture
-//!
-//! ### Parser
-//!
-//! The [`parser`] module provides the Pest-based parser with a trait abstraction
-//! for extensibility. The factory function [`parser::get_parser()`] returns a
-//! parser instance.
 //!
 //! ### Grammar File
 //!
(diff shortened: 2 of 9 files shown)


DominicBurkart (commented):
[CRON code review - session 3a0811bb]

**Scope**: Three intents (arch dedup, test coverage, clarity/doc fixes) across 8 files, +194/-29. Commits are labeled and each is independently reviewable per the description. Scope is on the high end of acceptable for a single PR, but the changes are small, orthogonal, and low-risk — splitting would add overhead without much benefit. Not recommending a split.

**Correctness concerns**

1. `marigold-impl/src/multi_consumer_stream.rs` — `MultiConsumerStream::run` spawns its pump via `crate::async_runtime::spawn` and returns immediately. The new tests (`single_consumer_receives_all_items`, `empty_source_closes_receiver`) do `mcs.run().await` and then `receiver.collect().await`. They pass only because the mpsc channel buffers and the spawned task eventually disconnects, but `run().await` is NOT a join. This tests observable behavior, which is fine — but the doc comment "run the pump to completion" is not what `run` does. Consider either (a) renaming the test to reflect that it exercises the public contract rather than any internal invariant, or (b) adding a `tokio::task::yield_now()`/`timeout` wrapper so a regression that made the pump hang would actually fail the test rather than hang CI.

2. `multi_consumer_stream.rs` test `run_future_as_stream_yields_no_items` — `RunFutureAsStream::new` takes `Pin<Box<F>>`, which `Box::pin(async { () })` produces, so this compiles. But the test only exercises the immediate-ready path; it never polls a `Pending` future, so it doesn't cover the `Poll::Pending` arm of `poll_next`. Worth adding.

3. `async_runtime.rs` dedup — changing the tokio arm's cfg to `#[cfg(feature = "tokio")]` is correct and the comment clearly documents the tokio-takes-precedence behavior. Good cleanup.

**Tests vs. invariants**

- `type_aggregation.rs` tests are solid: they cover the Constant/Variable lattice join, the Known/Unknown/Enum short-circuit, and the empty-iter identity element. These are the real invariants.
- `marifold` tests cover empty, non-zero init, and single-item — good coverage of the fold identity and accumulator.
- `collect_and_apply_empty_stream` and `_transformation` are thin but adequate. The renamed `_with_nested_async` name is accurate.
- `run_stream` passthrough + passthrough_empty is fine; the rename from `combinations` is clearly correct (it never tested combinations).

**Docs vs. code**

- README codecov badge fix: verified the target org/repo now matches. Removing the `?token=` query param is correct for public repos.
- `marigold-grammar/src/lib.rs` doc examples: `marigold_parse` is a real `pub fn` at the crate root, so the new path is accurate. The old `parser::parse_marigold` did not exist as a public API — good fix. Examples remain `ignore`d, so they aren't compile-tested; consider `no_run` instead so at least the types are checked.
- `marifold.rs` portmanteau doc comment is accurate and helpful.

**Bloat**: None observed. Removing `it_works` is a real improvement.

**Overall**: LGTM with two minor nits (test determinism in `multi_consumer_stream`, Pending-arm coverage for `RunFutureAsStream`). No blocking issues.

---

PR #192, opened 2026-04-24
fix: use --no-gpg-sign instead of config workaround (closes #68)

### LINKED ISSUE #68: solve gpg miscconfiguration

DominicBurkart:
## Summary

Closes #68 — "solve gpg miscconfiguration". Cherries #159's Rust-side fix (pass `--no-gpg-sign` directly on `git commit` invocations) onto current main, and adds a regression-guard test.

## Fingerprint

```
job-id: QO4Td
oldest-issue-id: one_track#2
this-issue: marigold#68
supersedes: marigold#159
parent-sha: fba35752597aeeb8d316bba272d1c1ef5030766d
base-at-fork: 4bbe781e303dd0329a66722ea084b39afe7287a8
```

## Changes

1. Replaced `git config commit.gpgsign false` step with `--no-gpg-sign` on `git commit` in bisect tests (`marigold-grammar/tests/bisect_cardinality.rs`, `marigold-grammar/tests/bisect_complexity.rs`).
2. Added `marigold-grammar/tests/no_gpg_config_workaround.rs` regression guard — fails if the old config workaround is reintroduced into either bisect test.

## Deferred

`.github/workflows/badges.yaml:47` update (`git commit` → `git commit --no-gpg-sign`) requires `workflows` token scope which the CRON agent does not have. Filed as a human-applied follow-up — a repo admin can apply this one-line change directly.

## Feedback loops

CI runs `Tests` on every PR — the regression guard piggybacks on it; any future PR that reintroduces `commit.gpgsign` config will fail `cargo test -p marigold-grammar`.

## Test plan

- [x] `cargo fmt --check` (clean)
- [x] `cargo test -p marigold-grammar` (29 tests pass including the new regression guard)
- [x] `cargo clippy -p marigold-grammar --test no_gpg_config_workaround -- -D warnings` (clean)
- [ ] Pre-existing clippy `-D warnings` errors on `marigold/src/main.rs` and `marigold-grammar/src/pest_ast_builder.rs` exist on `main` prior to this PR and are out of scope.

Closes #68

https://claude.ai/code/session_01JoDLmeYFCnEEmFmXBV25qY

---
_Generated by [Claude Code](https://claude.ai/code/session_01JoDLmeYFCnEEmFmXBV25qY)_

### DIFF
--- a/marigold-grammar/tests/bisect_cardinality.rs
+++ b/marigold-grammar/tests/bisect_cardinality.rs
@@ -29,7 +29,7 @@ fn commit_hash(dir: &std::path::Path) -> String {
 fn write_and_commit(dir: &std::path::Path, content: &str, message: &str) -> String {
     std::fs::write(dir.join("program.marigold"), content).unwrap();
     git(dir, &["add", "program.marigold"]);
-    git(dir, &["commit", "-m", message]);
+    git(dir, &["commit", "--no-gpg-sign", "-m", message]);
     commit_hash(dir)
 }
 
@@ -47,7 +47,6 @@ fn test_bisect_detects_exact_to_bounded_regression() {
     git(dir, &["init"]);
     git(dir, &["config", "user.email", "test@test.com"]);
     git(dir, &["config", "user.name", "Test"]);
-    git(dir, &["config", "commit.gpgsign", "false"]);
 
     let commit_a = write_and_commit(
         dir,
--- a/marigold-grammar/tests/bisect_complexity.rs
+++ b/marigold-grammar/tests/bisect_complexity.rs
@@ -29,7 +29,7 @@ fn commit_hash(dir: &std::path::Path) -> String {
 fn write_and_commit(dir: &std::path::Path, content: &str, message: &str) -> String {
     std::fs::write(dir.join("program.marigold"), content).unwrap();
     git(dir, &["add", "program.marigold"]);
-    git(dir, &["commit", "-m", message]);
+    git(dir, &["commit", "--no-gpg-sign", "-m", message]);
     commit_hash(dir)
 }
 
@@ -47,7 +47,6 @@ fn test_git_bisect_detects_memory_regression() {
     git(dir, &["init"]);
     git(dir, &["config", "user.email", "test@test.com"]);
     git(dir, &["config", "user.name", "Test"]);
-    git(dir, &["config", "commit.gpgsign", "false"]);
 
     let commit_a = write_and_commit(
         dir,
--- a/marigold-grammar/tests/no_gpg_config_workaround.rs
+++ b/marigold-grammar/tests/no_gpg_config_workaround.rs
@@ -1,0 +1,14 @@
+//! Regression guard: issue #68.
+//! Fails if the `git config commit.gpgsign false` workaround is reintroduced.
+
+#[test]
+fn bisect_tests_do_not_configure_gpg() {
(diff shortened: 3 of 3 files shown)


DominicBurkart (commented):
## Code review (automated, agent)

**Verdict: looks good, non-blocking nits only.** The change correctly addresses issue #68 by passing `--no-gpg-sign` per-invocation on `git commit` in the bisect tests, removing the global `git config commit.gpgsign false` workaround. Scope is appropriate.

### Alignment with issue #68
Issue #68 calls for avoiding `git config commit.gpgsign false`. Both occurrences (`marigold-grammar/tests/bisect_cardinality.rs:50`, `marigold-grammar/tests/bisect_complexity.rs:50` on `main`) are removed and `--no-gpg-sign` is added at the single shared `write_and_commit` helper in each file (line 32 post-change). A code search across the repo for `gpgsign` returns only these two test files, so the in-tree code surface is fully covered.

### Security scope of skipping GPG signing
Correct and narrow:
- `--no-gpg-sign` is applied only to `git commit` invocations inside ephemeral `TempDir` repos created by tests; it does not mutate the user's global/system git config nor any production code path.
- The previous `git config commit.gpgsign false` (writing into the temp repo's `.git/config`) was equally test-scoped, so this is not a security regression — if anything it is an improvement, because no persistent config state is written and the flag is local to each commit.
- No signing is bypassed in `publish.sh`, `install.sh`, or any release pipeline. `.github/workflows/badges.yaml:47` still does an unsigned `git commit -m "update badges"` but that pre-exists this PR and is an automation push by the badges workflow, not a developer/release commit.

### Regression guard
`marigold-grammar/tests/no_gpg_config_workaround.rs` uses `include_str!` + a substring check on the two bisect files. This is fine and CI-cheap, but two small caveats for future maintainers:
1. The substring is `"commit.gpgsign"` — if a future bisect test ever legitimately mentioned that string (e.g., a comment "do not reintroduce commit.gpgsign here"), the guard would false-positive. Low risk; acceptable.
2. The guard hard-codes the two filenames. If a third bisect test is added (e.g., `bisect_<x>.rs`), the guard will silently not cover it. A follow-up could glob `marigold-grammar/tests/bisect_*.rs` at test time via `std::fs::read_dir`. Not blocking.

### Deferred follow-up (badges.yaml)
The PR description correctly notes `.github/workflows/badges.yaml:47` (`git commit -m "update badges" || true`) cannot be updated by this CRON agent (lacks `workflows` scope). Confirmed the file on `main` still has the unsigned commit. A repo admin should apply `--no-gpg-sign` there for consistency, or the regression guard should be extended to cover workflow YAMLs once the workflows scope is available. Filing a tracking issue (if not already filed) is recommended.

### Bugs / bad trajectories / bloat
None found. Diff is +16/-4 across three files. No secrets/PII. No fabricated dependencies. `cargo fmt`/`cargo clippy` claims in the PR body are consistent with the diff.

### Tests / benches
- Added: regression guard test. Adequate for the stated goal.
- Suggested follow-up (not blocking): a positive test that sets `GIT_CONFIG_GLOBAL` (or `HOME`) to a temp dir containing `[commit] gpgsign = true` and verifies `write_and_commit` still succeeds. That would directly prove the original failure mode is fixed, rather than only that the workaround line is absent.

### Confidence
High (~0.9). Change is minimal, mechanically correct, and matches the linked issue. Only nit-level suggestions above.


---

PR #193, opened 2026-04-25
arch: dedupe runtime/bench code, drop redundant parse alias

DominicBurkart:
## Summary

Small architectural cleanups that remove duplication and simplify cfg-guarded code paths without changing public behavior.

- `marigold-impl/src/async_runtime.rs`: collapse three `spawn` cfg branches into two. The `tokio + async-std` branch was identical to the `tokio` branch and is now subsumed.
- `marigold-impl/src/multi_consumer_stream.rs`: hoist the duplicated body of `MultiConsumerStream::run` into a single `async move` block; only the spawn-vs-await wiring varies per cfg.
- `marigold-grammar/benches/parser_bench.rs`: remove 14 verbatim duplicate `c.bench_function(...)` registrations that re-registered the same benchmarks under the same names within each `bench_*` function.
- `marigold-grammar/src/lib.rs` + `marigold-macros/src/lib.rs`: drop the duplicate `marigold_grammar::marigold_parse` alias; `parser::parse_marigold` is the canonical entry point and the macro now imports it directly.

Net: 122 lines removed, 14 added; no public API removals beyond the unused `marigold_parse` alias whose only caller (the proc-macro crate) is updated in the same commit.

## Test plan

- [ ] `cargo build --workspace --all-features`
- [ ] `cargo test --workspace --all-features` (CLI install/uninstall tests rely on cargo network access; verified locally without)
- [ ] `cargo clippy --all-features --workspace -- -D warnings`
- [ ] `cargo fmt --all -- --check`

https://claude.ai/code/session_01FzKpNrStguLtayd2ia2SKq

---
_Generated by [Claude Code](https://claude.ai/code/session_01FzKpNrStguLtayd2ia2SKq)_

PR #194, opened 2026-04-25
test: cover MultiConsumerStream, RunFutureAsStream, type aggregation, run_stream

DominicBurkart:
## Summary

Adds tests to four previously zero- or single-test modules.

- `marigold-impl/src/multi_consumer_stream.rs` had no tests at all. Added six tests covering single-consumer pass-through, broadcast to multiple consumers, no-consumers completion, empty-inner termination, the `RunFutureAsStream` adapter (terminates yielding nothing), and its `size_hint` contract.
- `marigold-grammar/src/type_aggregation.rs` had no tests. Added seven unit tests covering the empty-iterator base cases, all-constant pass-through, any-variable propagation for `aggregate_input_variability`, and the empty/sum/Unknown-short-circuit/unresolved-Enum-short-circuit cases for `aggregate_input_count`.
- `marigold-impl/src/run_stream.rs` had a single happy-path test. Added long-stream order-preservation and empty-stream termination cases.

`InputCount`/`InputVariability` do not derive `Debug`, so the `assert!(... == ...)` form is used instead of `assert_eq!` to avoid widening the public type surface for tests alone.

## Test plan

- [x] `cargo test --workspace --exclude marigold --features tokio` — passes locally with all new tests
- [x] `cargo clippy --all-features --workspace -- -D warnings` — passes locally
- [x] `cargo fmt --all -- --check` — passes locally

## CI status

Two pre-existing failures appear on this PR and on the sibling arch PR (#193), neither of which touch the relevant code paths:

- `Tests` (workflow `tests.yaml`, `cargo test` job): also red on the arch PR which only edits `marigold-impl` and `marigold-grammar` benches/lib. Investigation required upstream — the failure occurs in <70s, suggesting an infrastructure/cargo issue rather than a test regression. The newly added tests in this PR all pass locally with `cargo test --features tokio`.
- `style hooks`: also red on the arch PR. The workflow runs `cargo audit && cargo deny check` plus `markdownlint`; advisory feeds drift independently of source changes.

Coverage delta: not measured (tarpaulin not installable in this sandbox). The new tests add coverage to two previously zero-test modules and extend a third.

https://claude.ai/code/session_01FzKpNrStguLtayd2ia2SKq

### DIFF
--- a/marigold-grammar/src/type_aggregation.rs
+++ b/marigold-grammar/src/type_aggregation.rs
@@ -26,3 +26,78 @@ pub fn aggregate_input_count<I: IntoIterator<Item = nodes::InputCount>>(
     }
     nodes::InputCount::Known(total_count)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::nodes::{InputCount, InputVariability};
+    use num_bigint::BigUint;
+
+    #[test]
+    fn variability_empty_is_constant() {
+        assert!(
+            aggregate_input_variability(std::iter::empty::<InputVariability>())
+                == InputVariability::Constant
+        );
+    }
+
+    #[test]
+    fn variability_all_constant_is_constant() {
+        assert!(
+            aggregate_input_variability(vec![
+                InputVariability::Constant,
+                InputVariability::Constant,
+                InputVariability::Constant,
+            ]) == InputVariability::Constant
+        );
+    }
+
+    #[test]
+    fn variability_any_variable_is_variable() {
+        assert!(
+            aggregate_input_variability(vec![
+                InputVariability::Constant,
+                InputVariability::Variable,
+                InputVariability::Constant,
+            ]) == InputVariability::Variable
+        );
+    }
+
+    #[test]
+    fn count_empty_is_known_zero() {
+        let result = aggregate_input_count(std::iter::empty::<InputCount>());
+        assert!(result == InputCount::Known(BigUint::from(0_u32)));
+    }
+
+    #[test]
+    fn count_sums_known_values() {
+        let result = aggregate_input_count(vec![
+            InputCount::Known(BigUint::from(2_u32)),
+            InputCount::Known(BigUint::from(3_u32)),
+            InputCount::Known(BigUint::from(7_u32)),
+        ]);
+        assert!(result == InputCount::Known(BigUint::from(12_u32)));
+    }
+
+    /// Encountering Unknown short-circuits regardless of any Known summands.
+    #[test]
+    fn count_unknown_propagates() {
+        let result = aggregate_input_count(vec![
+            InputCount::Known(BigUint::from(5_u32)),
+            InputCount::Unknown,
+            InputCount::Known(BigUint::from(10_u32)),
+        ]);
+        assert!(result == InputCount::Unknown);
+    }
+
+    /// An unresolved `Enum` placeholder behaves like Unknown for the purposes
+    /// of compile-time aggregation.
+    #[test]
+    fn count_unresolved_enum_propagates_as_unknown() {
+        let result = aggregate_input_count(vec![
+            InputCount::Known(BigUint::from(1_u32)),
+            InputCount::Enum("Words".to_string()),
+        ]);
+        assert!(result == InputCount::Unknown);
+    }
+}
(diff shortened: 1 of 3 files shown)


DominicBurkart (commented):
Good first coverage pass — the tests are readable, well-named, and cover the obvious happy-path and edge cases. A few observations, one of which is worth fixing before merge.

---

### 1. `multi_consumer_no_consumers_completes` tests the wrong thing (worth fixing)

`MultiConsumerStream::run()` is `async fn run(mut self)`. Under the `tokio` feature its body is:

```rust
pub async fn run(mut self) {
    self.senders.shrink_to_fit();

    crate::async_runtime::spawn(async move { … });   // ← spawns a new task, then returns immediately
}
```

So `run()` itself returns as soon as `async_runtime::spawn` is called — the actual producer loop runs inside the *newly-spawned* inner task. The test does:

```rust
let handle = tokio::spawn(mcs.run());
handle.await.expect("run() should complete cleanly");
```

`handle.await` resolves when the *outer* `run()` future completes, which happens almost instantly (right after the inner spawn). It does **not** wait for the inner producer task to drain the stream. The assertion therefore passes vacuously in all circumstances — it cannot catch a panic or error inside the producer loop.

A more meaningful variant: register a consumer and drive it to completion, which is what the single-consumer and empty-stream tests already do implicitly. If you want a standalone "no-consumer" test, you could at minimum document the caveat, or use a `JoinSet` / a small sleep-free synchronisation primitive to wait for the inner task.

---

### 2. `assert!` failure messages are silent on mismatch (minor, non-blocking)

For `InputCount` and `InputVariability` (no `Debug`), the `assert!(lhs == rhs)` form is correctly used — good. On failure though, the output is just `assertion failed` with no indication of which variant was actually returned. A custom message costs nothing:

```rust
assert!(
    result == InputCount::Unknown,
    "expected InputCount::Unknown, but the variants did not match"
);
```

Not a blocking issue, but it would make CI failures easier to diagnose.

---

### 3. `run_future_as_stream_size_hint_is_zero_unbounded` — `(0, None)` upper-bound semantics

The implementation unconditionally returns `(0, None)`. The test asserts exactly that, which is correct. One small note: `size_hint` returning `None` for the upper bound conventionally means "may produce infinitely many items", whereas `RunFutureAsStream` is guaranteed to produce *zero* items. A tighter (and still valid) hint would be `(0, Some(0))`. The current `(0, None)` is not *wrong* per the `Stream`/`Iterator` contract (implementations are permitted to be conservative), but callers that rely on the upper bound for pre-allocation would allocate unnecessarily. Whether to tighten the impl is a separate decision; the test correctly documents the *current* behaviour either way.

---

### 4. All async tests hard-code the `tokio` runtime

`#[tokio::test]` implicitly requires the `tokio` feature. The dev-dependency in `marigold-impl/Cargo.toml` pulls in `tokio = { version = "1", features = ["full"] }` unconditionally, so `cargo test` (without `--features tokio`) compiles and links tokio but the `multi_consumer_stream` production code takes the `#[cfg(not(any(feature = "async-std", feature = "tokio")))]` branch — meaning the tests exercise the non-feature production path while the `#[tokio::test]` harness is still active. This is an existing project-wide pattern (the test plan explicitly uses `--features tokio`) and is not introduced by this PR, but it is worth noting that the `async-std` feature path remains untested.

---

### Summary

| Finding | Severity |
|---|---|
| `multi_consumer_no_consumers_completes` awaits the wrong future | worth fixing before merge |
| `assert!` without custom messages | minor / non-blocking |
| `size_hint` returns `(0, None)` instead of `(0, Some(0))` | impl question, test is correct as-is |
| `async-std` path not covered | pre-existing gap, out of scope |

The `type_aggregation` and `run_stream` additions look correct and complete for the stated scope. The main ask is revisiting the no-consumers test so it actually exercises what its name promises.

https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX

---

DominicBurkart:
Review posted with one issue worth fixing before merge: `multi_consumer_no_consumers_completes` awaits a `JoinHandle` that resolves when the inner `tokio::spawn` call returns (not when the producer task finishes), making the test vacuous — it passes regardless of errors inside the producer. CI shows 3 pre-existing failures (`Tests`, `style hooks`, `Benches`) confirmed to affect every marigold PR on the current base — these are infra-level blockers unrelated to this PR's content, and must be resolved at the repo level before any marigold PR can merge. The vacuous test should be fixed, and the repo-level CI failures resolved, before opening for human review. [Session](https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX)

---

PR #205, opened 2026-04-26
fix(marigold): remove const from generated fn declarations — const fn restrictions break non-trivial user functions

DominicBurkart:
## Summary

- `FnDeclarationNode::code()` in `marigold-grammar/src/nodes.rs` was emitting `const fn {name}(...)` for every user-defined function.
- `const fn` in Rust prohibits heap allocations, closures, most trait method calls, and async code — so any non-trivial user function body would produce a confusing compile error in the generated code.
- The marigold architecture makes no claim that user functions must be `const`, so this was an accidental over-restriction.
- Fix: changed the single `format!` call to emit `fn {name}(...)` instead of `const fn {name}(...)`.

## Test plan

- [ ] Verify `marigold-grammar` unit tests still pass (`cargo test -p marigold-grammar`)
- [ ] Write a marigold program with a user-defined function that performs a heap allocation (e.g. returns a `Vec` or `String`) and confirm it compiles without error on this branch
- [ ] Confirm trivial functions (no heap allocs, no closures) still compile correctly

https://claude.ai/code/session_011LYFrHwghQBPNBWVeVhvjd

---
_Generated by [Claude Code](https://claude.ai/code/session_011LYFrHwghQBPNBWVeVhvjd)_

### DIFF
--- a/marigold-grammar/src/nodes.rs
+++ b/marigold-grammar/src/nodes.rs
@@ -734,7 +734,7 @@ impl FnDeclarationNode {
         let output_type = &self.output_type;
         let body = &self.body;
 
-        format!("const fn {name}({parameters_string}) -> {output_type} {{{body}}}")
+        format!("fn {name}({parameters_string}) -> {output_type} {{{body}}}")
     }
 }
 


DominicBurkart (commented):
## Review summary (changes requested — posted as COMMENT because review identity matches PR author)

The one-line change in `marigold-grammar/src/nodes.rs` (`const fn` -> `fn`) is correct and well-motivated: there is no architectural promise that user-defined functions must be `const`-eval safe, and `const fn` does forbid heap allocations, closures, most trait method calls, and `async`, which would silently break any non-trivial user function. The diff itself is minimal and sound.

However, **the PR as currently submitted breaks the build and is incomplete** relative to its own test plan.

### Blocker 1: existing unit tests still assert `const fn` and now fail

`marigold-grammar/src/parser.rs` contains four assertions that pin the old `const fn` output and were not updated:

- `parser.rs:1683` — `assert!(output.contains("const fn double"));` in `test_fn_simple_body`
- `parser.rs:1772` — `assert!(output.contains("const fn first"));` in `test_fn_multiple_functions`
- `parser.rs:1773` — `assert!(output.contains("const fn second"));` in `test_fn_multiple_functions`
- `parser.rs:1824` — `assert!(output.contains("const fn double"));` in `test_fn_with_stream`

Running `cargo test -p marigold-grammar --lib function_grammar_tests` on this branch produces:

```
failures:
    parser::function_grammar_tests::test_fn_multiple_functions
    parser::function_grammar_tests::test_fn_simple_body
    parser::function_grammar_tests::test_fn_with_stream

test result: FAILED. 22 passed; 3 failed; 0 ignored; 0 measured; 270 filtered out
```

The first checkbox of the PR's own test plan ("Verify `marigold-grammar` unit tests still pass") is therefore not satisfied. These assertions need to be updated to `assert!(output.contains("fn double"))` etc. Note: `"fn double"` is a substring of `"const fn double"`, so the new assertion alone is strictly weaker — recommend also asserting `!output.contains("const fn double")` to lock the regression in.

### Blocker 2: no regression test for the bug being fixed

The PR body says:

> Write a marigold program with a user-defined function that performs a heap allocation (e.g. returns a `Vec` or `String`) and confirm it compiles without error on this branch

…but no such test was added. Without one, this exact regression (regenerating `const fn` for user functions) can be reintroduced with no signal. Suggest adding a unit test in `function_grammar_tests` along the lines of:

```rust
#[test]
fn test_fn_with_heap_allocation_is_not_const() {
    let result =
        parse_marigold(r#"fn greet(name: String) -> String { format!("hi {}", name) }"#);
    assert!(result.is_ok());
    let output = result.unwrap();
    assert!(output.contains("fn greet"));
    assert!(!output.contains("const fn greet"));
}
```

…and ideally an end-to-end test under `marigold-grammar/tests/programs/` (mirroring the existing `multi_consumer.marigold` style) that compiles and runs a marigold program whose user fn returns a `String`/`Vec`.

### Other observations (non-blocking)

- **Single call site verified.** `FnDeclarationNode::code()` is only called from `parser.rs:186` (the user-fn emission path). There is no other place in the tree that emits `const fn ...` for user functions, so the one-line fix is complete in scope. No other `const` removal is needed.
- **No legitimate need for `const` here.** Marigold's generated user functions are invoked from async stream pipelines at runtime; nothing in the grammar, `marigold-impl`, or `marigold-macros` requires these to be evaluable in const context (no array-length usage, no `const`-generic instantiation, etc.). READMEs make no `const`/compile-time-eval promise (`grep -i const README.md marigold/README.md` returns nothing relevant). So no docs updates are required.
- **No perf regression.** Removing `const` only relaxes the compiler's checking; codegen of a non-const `fn` invoked at runtime is identical to that of a `const fn`, and the optimizer behavior (inlining etc.) is unchanged. No runtime cost.
- **Consider future-proofing.** If users ever want a truly const-evaluable helper (e.g. for array sizes), a follow-up could expose an explicit `const fn` form in the grammar (e.g. `const fn foo(...)`) rather than making it the default. Out of scope for this PR; just flagging.

### Requested changes before merge

1. Update the four `const fn` assertions in `marigold-grammar/src/parser.rs` (lines 1683, 1772, 1773, 1824) so the test suite passes.
2. Add at least one regression test that fails on `main` and passes here — minimum a unit test asserting the generated output does **not** contain `const fn` for a heap-allocating user function; ideally also a compiling end-to-end test exercising a `String`/`Vec`-returning user fn.
3. Re-run `cargo test -p marigold-grammar` and tick the first checkbox in the PR description.

Once those are in, this is good to merge.

_Automated review via [Claude Code](https://claude.ai/code/session_01WRjWCcDpY8waSwxaYtwcNw)_

---

PR #99, opened 2026-03-24
fix: keep_first_n panic on n=0, add tests for marigold-impl

DominicBurkart:
## Summary

- **Fix panic in `keep_first_n(0, ...)`**: Both the tokio and non-tokio code paths called `unwrap()` on an empty heap when `n=0`, causing a panic. Added an early return guard.
- **Add 23 new tests to `marigold-impl`** (6 → 29 total):
  - `keep_first_n`: 10 unit tests covering n=0, n=1 (find-max), n > stream length, empty stream, descending output order, tie-breaking, sorted/reverse-sorted input, custom comparators, n = stream length. Plus 3 proptests encoding key invariants (result length, top-n correctness, descending order).
  - `multi_consumer_stream`: 6 tests (from 0) — single/dual consumer fan-out, empty stream, no-consumer edge case, `RunFutureAsStream` behavior and size_hint.
  - `writer`: 2 tests (from 0, `io` feature) — vector writer write+flush, file writer roundtrip.
- **Add `proptest` dev-dependency** to `marigold-impl`.

## Bug details

The proptest `result_length_is_min_of_n_and_input_len` discovered that `keep_first_n(0, ...)` panics at the `unwrap()` on line 108 (`first_n_mutex.lock().peek().unwrap()`). The heap is empty when n=0, so `peek()` returns `None`. The regression case is checked in via `proptest-regressions/keep_first_n.txt`.

## Test plan

- [x] `cargo test --features tokio -p marigold-impl` — 27 passed
- [x] `cargo test --features tokio,io -p marigold-impl` — 29 passed
- [x] `cargo test --all-features` — full workspace green
- [x] `cargo clippy --all-features -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

https://claude.ai/code/session_01JAjtBEp74RC1LvLjNm2jm8

PR #214, opened 2026-04-29
feat(marigold-impl): compile-time error on keep_first_n(0) (closes #211)

### LINKED ISSUE #211: compile-time error on keep_first_n where n == 0
If n is a static value, I don't see a semantic reason: it's always a noop so the input stream is never polled. It theoretically could be the value of a function, but that's not valid syntax today, so we can defer worrying about it.

Provide a helpful, descriptive error at compile time for this case.

cf https://github.com/DominicBurkart/marigold/pull/99/changes#diff-12e16b65ceadf6817902f6da64c9e46f605c20b0e1a1a87c545d35e3a6800cdfL201

DominicBurkart:
## Why

Closes #211. With a static literal `n == 0`, `keep_first_n(0, …)` is provably a no-op — the input stream is never polled, no items are ever emitted. Today's grammar only allows numeric literals at this position, so the literal-zero case is unambiguously detectable at macro-parse time and almost certainly a bug. The issue (linking to PR #99) asked for a helpful, descriptive compile-time error in this case.

(A function-valued `n` is not yet expressible in the grammar, so deferring that case as the issue suggests.)

## What changed

- `marigold-grammar/src/pest_ast_builder.rs`: in `build_stream_function`, the `Rule::keep_first_n_fn` arm now checks the literal extracted by `peek_numeric_arg`. If `n == 0` it returns `Err(...)` with a descriptive message instead of lowering. The existing `marigold!`/`m!` proc macro path turns this into a `proc macro panicked` diagnostic with the message attached as `help: message: ...` — same shape as the existing bounded-type validation errors.
- `marigold-grammar/src/parser.rs`: two new unit tests:
  - `test_reject_keep_first_n_zero_literal` — confirms the parser surfaces an error mentioning `keep_first_n(0)` and `no-op`.
  - `test_accept_keep_first_n_one_literal` — positive smoke test that `keep_first_n(1, …)` continues to lower correctly.
- `examples/bounded-types/compile_fail/keep_first_n_zero.{rs,stderr}`: end-to-end trybuild fixture exercising the diagnostic through `m!()`.

## Diagnostic snippet (trybuild .stderr)

```
error: proc macro panicked
 --> compile_fail/keep_first_n_zero.rs:6:13
  |
6 |       let _ = m!(
  |  _____________^
7 | |         range(0, 10).keep_first_n(0, compare).return
8 | |     )
  | |_____^
  |
  = help: message: marigold parsing error: MarigoldParseError("keep_first_n(0) is a no-op — the input stream is never polled. Remove the call or pass a non-zero literal.")
```

## Test results

- `cargo test -p marigold-grammar --lib keep_first_n` → 5 passed (including the two new tests).
- `cargo test -p marigold-impl` → all green (28/4/5/7/4/8/0). The runtime `keep_first_n_zero` integration test still passes since this PR only rejects at proc-macro parse time; direct `KeepFirstN` trait callers are unaffected.
- `cargo test -p marigold` → all green (4 passed in long-running suite, plus 0/1 in unit/doc).
- `cargo fmt --check` → clean.

## Known pre-existing breakage (NOT introduced here)

- `cargo clippy --all-targets -- -D warnings` fails on `origin/main` with 4 `unnecessary_map_or` errors in `marigold-grammar/tests/proptest_analyze.rs` and a `dead_code` error on `split_respecting_parens`. Reproduced cleanly on `origin/main` with my changes stashed. Out of scope for #211.
- The 4 pre-existing `examples/bounded-types/compile_fail/*.stderr` fixtures are already stale on `origin/main` due to a Rust toolchain bump that now appends an `E0282` cascade after the proc-macro panic. Reproduced on bare main. Out of scope for #211 — PR #184 ("fix(ci): Rust 1.95.0 compatibility on main") looks like the right venue. The new `keep_first_n_zero.stderr` fixture in this PR includes the `E0282` cascade so it matches the current toolchain output.

## Trailers

- job-id: No3dF
- oldest-issue-id: 211

---
_Generated by [Claude Code](https://claude.ai/code/session_01J1adCZuguG5uuoeQjR3oh8)_

### DIFF
--- a/examples/bounded-types/compile_fail/keep_first_n_zero.stderr
+++ b/examples/bounded-types/compile_fail/keep_first_n_zero.stderr
@@ -1,0 +1,20 @@
+error: proc macro panicked
+ --> compile_fail/keep_first_n_zero.rs:6:13
+  |
+6 |       let _ = m!(
+  |  _____________^
+7 | |         range(0, 10).keep_first_n(0, compare).return
+8 | |     )
+  | |_____^
+  |
+  = help: message: marigold parsing error: MarigoldParseError("keep_first_n(0) is a no-op — the input stream is never polled. Remove the call or pass a non-zero literal.")
+
+error[E0282]: type annotations needed
+ --> compile_fail/keep_first_n_zero.rs:6:13
+  |
+6 |       let _ = m!(
+  |  _____________^
+7 | |         range(0, 10).keep_first_n(0, compare).return
+8 | |     )
+9 | |     .await
+  | |__________^ cannot infer type
--- a/marigold-grammar/src/parser.rs
+++ b/marigold-grammar/src/parser.rs
@@ -1258,6 +1258,40 @@ mod negative_tests {
     }
 
     #[test]
+    fn test_reject_keep_first_n_zero_literal() {
+        // Issue #211: a static keep_first_n(0) is a no-op (the input stream is
+        // never polled), so the macro should reject it at compile time with a
+        // helpful diagnostic rather than emitting silently dead code.
+        let result = parse_marigold("range(0, 10).keep_first_n(0, compare).return");
+        assert!(
+            result.is_err(),
+            "keep_first_n(0) should be rejected at parse time (no-op)"
+        );
+        let err = result.unwrap_err().0;
+        assert!(
+            err.contains("keep_first_n(0)") && err.contains("no-op"),
+            "Error should mention keep_first_n(0) and no-op, got: {err}"
+        );
+    }
+
+    #[test]
+    fn test_accept_keep_first_n_one_literal() {
+        // Positive smoke test: keep_first_n with non-zero literal continues to
+        // parse and lower correctly after the n==0 guard was added.
+        let result = parse_marigold("range(0, 10).keep_first_n(1, compare).return");
+        assert!(
+            result.is_ok(),
+            "keep_first_n(1) should still parse: {:?}",
+            result
+        );
+        let code = result.unwrap();
+        assert!(
+            code.contains("keep_first_n(1, compare)"),
+            "lowered code should contain keep_first_n(1, compare), got: {code}"
+        );
+    }
+
+    #[test]
     fn test_permutations_generates_array_conversion() {
         let code = parse_marigold("range(0, 5).permutations(2).return")
             .expect("Should parse permutations(2)");
--- a/marigold-grammar/src/parser.rs
+++ b/marigold-grammar/src/parser.rs
@@ -1275,6 +1275,25 @@ mod negative_tests {
     }
 
     #[test]
+    fn test_reject_keep_first_n_leading_zero_literal() {
+        // Issue #211 follow-up: the n==0 guard is on numeric value (post-`u64` parse),
+        // not lexical form. Lock that in by exercising a leading-zero literal `00` —
+        // this should be rejected just like `0`, so a future grammar/parser refactor
+        // (e.g. switching to a different numeric token shape) cannot silently regress
+        // this check.
+        let result = parse_marigold("range(0, 10).keep_first_n(00, compare).return");
+        assert!(
+            result.is_err(),
+            "keep_first_n(00) should be rejected at parse time (no-op)"
+        );
+        let err = result.unwrap_err().0;
+        assert!(
+            err.contains("keep_first_n(0)") && err.contains("no-op"),
+            "Error should mention keep_first_n(0) and no-op, got: {err}"
+        );
+    }
+
+    #[test]
     fn test_accept_keep_first_n_one_literal() {
         // Positive smoke test: keep_first_n with non-zero literal continues to
         // parse and lower correctly after the n==0 guard was added.
--- a/marigold-grammar/src/pest_ast_builder.rs
+++ b/marigold-grammar/src/pest_ast_builder.rs
@@ -856,6 +856,13 @@ impl PestAstBuilder {
             }
             Rule::keep_first_n_fn => {
                 let n = Self::peek_numeric_arg(&inner)?;
+                if n == 0 {
+                    return Err(
+                        "keep_first_n(0) is a no-op — the input stream is never polled. \
+                         Remove the call or pass a non-zero literal."
+                            .to_string(),
+                    );
+                }
                 (
                     StreamFunctionKind::KeepFirstN(n),
                     Self::build_keep_first_n_fn(inner)?,
--- a/marigold-grammar/src/pest_ast_builder.rs
+++ b/marigold-grammar/src/pest_ast_builder.rs
@@ -855,6 +855,13 @@ impl PestAstBuilder {
                 )
             }
             Rule::keep_first_n_fn => {
+                // Literal-only guard: today's grammar admits `free_text_literal = ASCII_DIGIT+`
+                // at this position, so `peek_numeric_arg` parses a `u64` from the source text.
+                // That means `n == 0` covers `0`, `00`, `0_000`, etc. — any lexical form whose
+                // numeric value is zero. Function-valued / const-generic `n` is explicitly
+                // deferred per issue #211 and is not yet expressible in the grammar; if/when
+                // the grammar grows to admit non-literal arguments here, this guard must be
+                // revisited (it would silently become a partial check).
                 let n = Self::peek_numeric_arg(&inner)?;
                 if n == 0 {
                     return Err(
--- a/examples/bounded-types/compile_fail/keep_first_n_zero.rs
+++ b/examples/bounded-types/compile_fail/keep_first_n_zero.rs
@@ -1,0 +1,16 @@
+use marigold::m;
+use marigold::marigold_impl::StreamExt;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let _ = m!(
+        range(0, 10).keep_first_n(0, compare).return
+    )
+    .await
+    .collect::<Vec<()>>()
+    .await;
+}
+
+fn compare(a: &i32, b: &i32) -> std::cmp::Ordering {
+    a.cmp(b)
+}
--- a/examples/bounded-types/compile_fail/min_greater_than_max.stderr
+++ b/examples/bounded-types/compile_fail/min_greater_than_max.stderr
@@ -10,3 +10,14 @@ error: proc macro panicked
    | |_____^
    |
    = help: message: marigold parsing error: MarigoldParseError("Bounded type validation failed:\n  - Bounds violation for field 'Sensor.reading': min (100) is greater than max (0)")
+
+error[E0282]: type annotations needed
+  --> compile_fail/min_greater_than_max.rs:6:13
+   |
+ 6 |       let _ = m!(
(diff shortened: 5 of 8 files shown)


marigold-grammar/src/pest_ast_builder.rs
```diff
@@ -856,6 +856,13 @@ impl PestAstBuilder {
             }
             Rule::keep_first_n_fn => {
                 let n = Self::peek_numeric_arg(&inner)?;
+                if n == 0 {
+                    return Err(
+                        "keep_first_n(0) is a no-op — the input stream is never polled. \
+                         Remove the call or pass a non-zero literal."
+                            .to_string(),
```
DominicBurkart, line 870: nit (non-blocking): the guard correctly catches `0`, `00`, `0_000` etc. since `peek_numeric_arg` does `parse::<u64>()` after `free_text_literal = ASCII_DIGIT+`. Worth a one-line code comment noting *why* this is sound (i.e. grammar today only admits literal digits, so a function-valued n is impossible — explicitly the deferred case from #211). That makes future-grammar drift obvious to the next reader.

---

examples/bounded-types/compile_fail/keep_first_n_zero.stderr
```diff
@@ -0,0 +1,20 @@
...
+
+error[E0282]: type annotations needed
+ --> compile_fail/keep_first_n_zero.rs:6:13
+  |
+6 |       let _ = m!(
+  |  _____________^
+7 | |         range(0, 10).keep_first_n(0, compare).return
+8 | |     )
+9 | |     .await
+  | |__________^ cannot infer type
```
DominicBurkart, line 20: nit (non-blocking): the `E0282` cascade is toolchain-version dependent. PR body acknowledges this and other sibling fixtures already drift on toolchain bumps. Consider tracking a follow-up to either (a) bind/annotate the type in the trybuild fixture so E0282 doesn't fire, or (b) refactor compile_fail fixtures to a more stable shape — but out of scope here.

---

examples/bounded-types/compile_fail/keep_first_n_zero.stderr
```diff
@@ -0,0 +1,20 @@
+error: proc macro panicked
+ --> compile_fail/keep_first_n_zero.rs:6:13
+  |
+6 |       let _ = m!(
+  |  _____________^
+7 | |         range(0, 10).keep_first_n(0, compare).return
+8 | |     )
+  | |_____^
+  |
+  = help: message: marigold parsing error: MarigoldParseError("keep_first_n(0) is a no-op — the input stream is never polled. Remove the call or pass a non-zero literal.")
```
DominicBurkart, line 10: nit: surface diagnostic is `proc macro panicked` + `help: message: marigold parsing error: MarigoldParseError("...")` — the user-visible string includes the `MarigoldParseError(...)` debug wrapper and escaped quotes, which is uglier than a plain `error: keep_first_n(0) is a no-op...`. This matches the existing `MarigoldParseError`-based path so it's consistent with the rest of the macro's diagnostics, but a future cleanup could route these through a real `proc_macro::Diagnostic` / `compile_error!` for a cleaner UX. Non-blocking.

---

marigold-grammar/src/parser.rs
```diff
@@ -1257,6 +1257,40 @@ mod negative_tests {
...
+        assert!(
+            result.is_ok(),
+            "keep_first_n(1) should still parse: {:?}",
+            result
+        );
+        let code = result.unwrap();
+        assert!(
+            code.contains("keep_first_n(1, compare)"),
+            "lowered code should contain keep_first_n(1, compare), got: {code}"
+        );
```
DominicBurkart, line 1310: Optional: consider adding a parser-level negative test for `keep_first_n(00, compare)` (or similar leading-zero literal) to lock in that the guard is on numeric value, not lexical form. Cheap and prevents a future grammar/parser refactor from regressing this silently. Non-blocking.

---

DominicBurkart (commented):
Reviewed (cannot self-approve; treat as APPROVE-equivalent, no blockers).

Scope matches issue #211 exactly: literal `n == 0` rejected, function-valued `n` deferred (and not currently expressible in the grammar anyway). The check sits in `build_stream_function`'s `Rule::keep_first_n_fn` arm and bubbles via `Err(String)` to the proc-macro entry point, so this *is* compile-time (proc-macro parse time), not runtime — confirmed by the trybuild fixture failing during compilation.

Test coverage is appropriate:
- `test_reject_keep_first_n_zero_literal` (negative parser test) asserts the message mentions `keep_first_n(0)` and `no-op`.
- `test_accept_keep_first_n_one_literal` (positive smoke test) confirms non-zero literals still lower correctly.
- `examples/bounded-types/compile_fail/keep_first_n_zero.{rs,stderr}` exercises end-to-end via `m!` and is auto-picked up by the existing `compile_fail/*.rs` trybuild glob in `bounded-types/src/main.rs`.
- Existing runtime test `marigold-impl/tests/keep_first_n_edge_cases.rs::keep_first_n_zero` is correctly unaffected since direct `KeepFirstN` callers bypass the macro.

Edge cases:
- Negative literals: grammar's `free_text_literal = ASCII_DIGIT+` precludes them; `peek_numeric_arg` returns `u64`. OK.
- `u64::MAX` and other large values: untouched.
- Leading-zero literals (`00`, `000`): the `n == 0` guard is on numeric value post-parse, so handled. Worth a parser test (see inline) but not blocking.
- Const-generic / function-valued n: explicitly deferred per issue.

Scope is tight (4 files, +77 lines), no scope creep, no secrets/PII, no clippy/fmt regressions introduced (pre-existing breakage on main acknowledged and out of scope).

Non-blocking nits inline: (1) brief comment explaining why grammar makes the literal-only assumption sound, (2) consider parser test for `keep_first_n(00, ...)`, (3) diagnostic surface wraps the message in `MarigoldParseError(...)` debug — consistent with siblings but a future UX cleanup target, (4) `E0282` cascade in the .stderr is toolchain-brittle as noted in PR body.

---

marigold-grammar/src/pest_ast_builder.rs
DominicBurkart: Addressed in a95eabe — added a 7-line block comment in the `Rule::keep_first_n_fn` arm of `build_stream_function` explaining (a) why the literal-only assumption is sound today (`free_text_literal = ASCII_DIGIT+` so `peek_numeric_arg` always parses a `u64` and `n == 0` covers `0`/`00`/`0_000`), and (b) that this guard must be revisited if the grammar grows to admit non-literal `n` here (deferred case from #211).

---

examples/bounded-types/compile_fail/keep_first_n_zero.stderr
DominicBurkart: Acknowledged — addressed in 8d94a68 (regenerated stale trybuild `.stderr` fixtures so the new fixture matches sibling drift on the current toolchain). Tracking the structural fix (bind/annotate the type so `E0282` doesn't fire, or refactor compile_fail fixture shape) as a follow-up; out of scope for #211.

---

examples/bounded-types/compile_fail/keep_first_n_zero.stderr
DominicBurkart: Deferred — kept consistent with the existing `MarigoldParseError`-based diagnostic shape used by every other bounded-type validation error in the macro. Routing these through `proc_macro::Diagnostic` / `compile_error!` for cleaner UX is a worthwhile cross-cutting cleanup but out of scope for #211; happy to file a follow-up issue if useful.

---

marigold-grammar/src/parser.rs
DominicBurkart: Addressed in a95eabe — added `test_reject_keep_first_n_leading_zero_literal` (in `marigold-grammar/src/parser.rs`, alongside the existing `test_reject_keep_first_n_zero_literal`) which exercises `keep_first_n(00, compare)` and asserts the error mentions `keep_first_n(0)` and `no-op`. Locks in the numeric-value (not lexical-form) semantics of the guard so a future grammar/parser refactor can't silently regress it.

---

DominicBurkart:
CRON status: code-review APPROVE-equivalent (compile-time enforcement is genuine — proc-macro parse-time `Err`, not runtime; tight scope at +77/4 files) with 4 non-blocking inline nits. Address-comments shipped a95eabe (literal-only block comment in `pest_ast_builder.rs::Rule::keep_first_n_fn` arm + new parser test `test_reject_keep_first_n_leading_zero_literal`) and replied to all 4 threads; prior CRON had also pushed 8d94a68 regenerating trybuild `.stderr` fixtures for the toolchain's E0282 cascade. `cargo fmt --all -- --check` + `cargo test -p marigold-impl` + `cargo test -p marigold-grammar --lib keep_first_n` (6 pass) all green; pending CI re-run. https://claude.ai/code/session_016Z4LEfbFtcxqDmSanu3zFW

---

DominicBurkart:
CRON coordination ping — `job-id: cron-2026-05-07-QPJmn` was tasked with #211 but PR #214 (`job-id: No3dF`, < 14d) already has the implementation: literal-zero guard in `pest_ast_builder.rs::Rule::keep_first_n_fn`, parser unit tests (`test_reject_keep_first_n_zero_literal`, `test_reject_keep_first_n_leading_zero_literal`), and the trybuild fixture under `examples/bounded-types/compile_fail/keep_first_n_zero.{rs,stderr}`. Verified locally: `cargo build -p marigold-grammar` clean against the head SHA (`a95eabe`).

Heads-up though: the **2026-05-07T11:40Z** check-runs on `a95eabe` (same SHA as the prior green run on 2026-04-30) flipped to failure across Tests / all platform builds / style hooks / Benches without a new commit. Likely a re-trigger picking up new toolchain/dep drift on main rather than a regression introduced by this PR. Worth a glance — if it's the trybuild `E0282` cascade drifting again, the fix lives in the `.stderr` fixtures touched by `8d94a68`.

Not duplicating; stopping here on #211 to avoid 3-way collision.

job-id: cron-2026-05-07-QPJmn
oldest-issue-id: 211

---

DominicBurkart:
CRON status: NO-OP. Fingerprint match against head `a95eabe6b2121da868a1e57c0fc2d5484eca5461` — implementation, tests, and trybuild fixture for #211 are all in place; all 4 owner review threads addressed by `a95eabe` (literal-only block comment in `pest_ast_builder.rs::Rule::keep_first_n_fn`, `test_reject_keep_first_n_leading_zero_literal` parser test) with `8d94a68` regenerating sibling `.stderr` fixtures for the E0282 cascade. No new commits since prior CRON's address-comments pass; no unaddressed feedback; no intervening pushes to `main` per `git log` of branch base. Honoring `cron-2026-05-07-QPJmn`'s collision-avoidance — not duplicating work. CI status on `a95eabe` remains `pending` after the 2026-05-07T11:40Z re-trigger flipped Tests/Benches; per prior diagnosis those failures track toolchain drift on `main` (the same shape addressed by `8d94a68`'s fixture regen), not a regression in this PR's scope.

[job:RlQo0][oldest-issue:#211]

---

PR #221, opened 2026-05-01
chore: maintenance pass (architecture, tests, clarity)

DominicBurkart:
Maintenance pass against `main`. All workstreams are unrelated to feature work; each commit targets one logical change.

## 1. Architecture alignment & simplification

- **`marigold-grammar/src/lib.rs`**: dropped the placeholder `#[test] fn it_works()` (2+2 sanity check, no real coverage) and the unused `extern crate proc_macro;` (the grammar crate exposes no proc-macros — only `marigold-macros` does). See `marigold-grammar/src/lib.rs:103-108` and `marigold-grammar/src/lib.rs:65` (pre-change).
- **`marigold-grammar/src/bound_expr.rs` (deleted, ~680 LOC)**: the file was never declared as a module in `marigold-grammar/src/lib.rs` and had no consumers; the actual bound-expression parsing is provided by the Pest grammar rule `bound_expr` (see `marigold-grammar/src/parser.rs:635-672`).
- **`examples/bounded-types/compile_fail/*.stderr` (4 files)**: regenerated via `TRYBUILD=overwrite`. Modern stable rustc emits an additional `error[E0282]: type annotations needed` cascade after each proc-macro panic; the fixtures lacked this block, so `cargo test --all-features` failed in `bounded_types::tests::compile_fail`. Diagnosed at `examples/bounded-types/src/main.rs:172-175`.

## 2. Testing posture

- **`marigold-impl/src/multi_consumer_stream.rs:104-165` (new)**: added four unit tests for a previously-untested module that is core glue for every named-stream marigold program.
  - Component: `MultiConsumerStream` + `RunFutureAsStream`.
  - Disposition before: zero inline tests, no integration tests.
  - Invariants validated:
    1. Every registered consumer receives every source item, in source order.
    2. `run()` drains the source even with zero registered receivers.
    3. Empty source yields empty receivers (terminates promptly).
    4. `RunFutureAsStream` yields zero items and ends after the wrapped future resolves.
- **`marigold-impl/src/combinations.rs` and `marigold-impl/src/permutations.rs`**: deduped inline tests against the integration suites in `marigold-impl/tests/`. The intra-module tests duplicated `k=0`, `k>n`, `k=n`, and empty-stream cases already covered by `tests/combinations_edge_cases.rs` and `tests/permutations_edge_cases.rs`. Kept only the inline tests asserting exact emission ordering and cardinality counts (which the integration suites do not). Net -76 LOC, no coverage loss.

## 3. Clarity

[...]

PR #223, opened 2026-05-01
test: add coverage for multi_consumer_stream and writer modules

DominicBurkart:
## Summary

- `marigold-impl/src/multi_consumer_stream.rs` and `marigold-impl/src/writer.rs` had **zero tests**, making them the largest coverage gap in the repo as measured by codecov
- Adds 8 new tests across two new test files to close that gap

### `marigold-impl/tests/multi_consumer_stream_tests.rs` (new, 4 tests)
- Single consumer receives all items from `MultiConsumerStream`
- Two consumers **both** receive every item (broadcast correctness) — uses `tokio::join!` to drain concurrently, avoiding the deadlock caused by the channel's buffer-size-1 constraint
- Empty-stream terminates cleanly (senders disconnect)
- `RunFutureAsStream` drives a future to completion but yields no stream items

### `marigold-impl/tests/writer_tests.rs` (new, 4 tests, gated on `feature = "io"`)
- `Writer::vector()` — write + flush + shutdown succeed without error
- `Writer::vector()` — multiple sequential writes succeed
- `Writer::file()` — write bytes, shutdown, then read back and verify contents match
- `Writer::file()` — large write (64 KB) succeeds end-to-end

## Test plan
- [ ] `cargo test -p marigold-impl --features "tokio io"` passes locally (40 pre-existing + 8 new = 48 tests green)
- [ ] CI passes on this PR

---
_Generated by [Claude Code](https://claude.ai/code/session_01KTC47SJvyUUxhjh66UU8LG)_

### DIFF
--- a/marigold-impl/Cargo.toml
+++ b/marigold-impl/Cargo.toml
@@ -42,6 +42,7 @@ tokio = { version = "1", features = ["full"]}
 criterion = { version = "0.5", features = ["html_reports"] }
 cpu-time = "1.0"
 tracing-subscriber = { version = "0.3", features = ["registry"] }
+tempfile = "3"
 
 [[bench]]
 name = "keep_first_n"
--- a/marigold-impl/src/writer.rs
+++ b/marigold-impl/src/writer.rs
@@ -25,6 +25,17 @@ impl Writer {
             inner: WriteTarget::Vector(Box::pin(Vec::new())),
         }
     }
+
+    /// Returns the underlying `Vec<u8>` for a `Writer::vector()` instance, or
+    /// `None` for other variants. Exposed primarily so tests and callers that
+    /// constructed the writer with [`Self::vector`] can inspect or take
+    /// ownership of the buffered bytes after writes complete.
+    pub fn into_vec(self) -> Option<Vec<u8>> {
+        match self.inner {
+            WriteTarget::Vector(boxed) => Some(*Pin::into_inner(boxed)),
+            WriteTarget::File(_) => None,
+        }
+    }
 }
 
 impl tokio::io::AsyncWrite for Writer {
--- a/marigold-impl/tests/multi_consumer_stream_tests.rs
+++ b/marigold-impl/tests/multi_consumer_stream_tests.rs
@@ -85,5 +85,8 @@ async fn run_future_as_stream_yields_no_items_but_runs_future() {
     let items: Vec<u32> = stream.collect().await;
 
     assert!(items.is_empty(), "RunFutureAsStream should yield no items");
-    assert!(flag.load(Ordering::SeqCst), "Future should have been driven to completion");
+    assert!(
+        flag.load(Ordering::SeqCst),
+        "Future should have been driven to completion"
+    );
 }
--- a/marigold-impl/tests/multi_consumer_stream_tests.rs
+++ b/marigold-impl/tests/multi_consumer_stream_tests.rs
@@ -1,32 +1,44 @@
-//! Tests for `MultiConsumerStream` and `RunFutureAsStream`.
+//! Tests for `MultiConsumerStream`.
+//!
+//! Scope note: PR #221 introduces inline unit tests in
+//! `marigold-impl/src/multi_consumer_stream.rs` that cover (a) the empty-source
+//! case, (b) `run()` with no receivers, and (c) `RunFutureAsStream`. To avoid
+//! duplicate coverage, this integration suite focuses on cases #221 does not
+//! exercise:
+//!
+//!  * a single-consumer happy path that exercises ordering end-to-end;
+//!  * a two-consumer broadcast that verifies both receivers see every item in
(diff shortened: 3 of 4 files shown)


DominicBurkart (commented):
### Review summary (for coding agents)

The PR closes a real coverage gap with mostly meaningful tests (assertions verify behavior, not just exercise code paths). However, several tests are weaker than they appear, and important concurrency/error edge cases are missing for two modules whose value is precisely in those edge cases. Production code is untouched. No property tests despite obvious candidates (broadcast invariants, write-then-read round-trip).

Key gaps:

1. **`MultiConsumerStream` tests only exercise the `tokio` spawn path.** `MultiConsumerStream::run()` has two cfg-gated implementations: when `tokio`/`async-std` is enabled it spawns a background task; otherwise it drains inline. The new tests await `mcs.run()` *before* reading any receiver. With `BUFFER_SIZE = 1`, the inline (no-runtime-feature) path will deadlock at the second `feed` because no receiver is being polled. Coverage of the `cfg(not(any(...)))` branch is therefore zero, and the test file does not declare `required-features` in `Cargo.toml` — this should be made explicit so the test isn't silently skipped or misleadingly run.

2. **No concurrency edge cases** for the headline module:
   - Slow consumer / fast consumer asymmetry (does a stalled receiver block the other?).
   - Fan-out fairness with N>2 consumers.
   - Consumer dropped mid-stream (does sender survive the broken `Sender::feed`?).
   - Source stream that yields `Pending` (the `iter` source is always Ready).
   - Backpressure: with `BUFFER_SIZE=1`, a slow consumer must observably gate the fast one — assert that ordering/blocking holds.

3. **`Writer::vector()` tests assert nothing about written bytes.** The `Vec` is an internal `Pin<Box<Vec<u8>>>` with no accessor, so the tests can only check that `write_all`/`flush`/`shutdown` return `Ok`. This is coverage padding rather than behavioral verification — once `Writer::vector()` exists at all, these tests are nearly tautological. Either expose a way to extract the buffer (production change) or drop the vector tests in favor of file-only coverage.

4. **No error-path / partial-write / backpressure coverage for `Writer`.** The `AsyncWrite` impl has three forwarding methods (`poll_write`, `poll_flush`, `poll_shutdown`) and two `WriteTarget` variants — six branches. Tests cover the happy path on each variant but never:
   - A failing underlying writer (e.g., file opened read-only, file dropped/closed before flush).
   - `poll_write` returning `Pending` then `Ready`.
   - Partial writes (`write` rather than `write_all`).
   - Behavior after `shutdown` (writes after shutdown should error).

5. **No property tests.** `MultiConsumerStream` has obvious invariants (`for_all consumers, received == source`); `Writer::file` has a write-then-read round-trip property. `proptest`/`quickcheck` would yield much stronger guarantees for the same surface area.

6. **Temp file hygiene.** `std::env::temp_dir().join(format!("..._{}.txt", std::process::id()))` is racy: re-running the test binary doesn't guarantee a fresh path if a prior run left files; concurrent invocations (e.g., codecov + main test job on the same runner) can collide. Use `tempfile::NamedTempFile` (already idiomatic in Rust integration tests) or include a nonce + `tokio::fs::remove_file` in a guard that runs on panic.

7. **`run_future_as_stream_yields_no_items_but_runs_future` is solid** — it asserts the side-effect flag and the empty-stream output, which is the right shape for this test.

No production code touched, no scaffolding bloat. Total ~170 LOC for 8 tests is reasonable, but a few of those tests are doing very little work and should either be strengthened or removed.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

DominicBurkart:
**Cron status — 3 subagents complete**

CR posted (concurrency gaps in `MultiConsumerStream` tests — only the tokio-spawned path was exercised; `Writer::vector()` tests verified nothing about contents; PID-based temp paths racy across runs; significant duplication with PR #221).

Address-comments pushed `f3b375c0` addressing 4 items: deduplicated overlap with #221 (kept the unique single-consumer-ordering + 2-consumer-broadcast cases #221 lacks); replaced `std::env::temp_dir()` + `std::process::id()` with `tempfile::TempDir`; added `Writer::into_vec(self) -> Option<Vec<u8>>` accessor + content-assertion tests (`vector_writer_buffers_concatenated_content`, `_empty_yields_empty_buffer`, `file_writer_into_vec_returns_none`); added `slow_consumer_does_not_lose_items_for_anyone` under `BUFFER_SIZE=1` backpressure with paced reads. Local: 66 tests green, fmt+clippy clean.

CI: `Tests` was cancelled (6h runner timeout, not code-induced); `Benches` pre-existing on main. Both depend on #221 landing first. Branch-name mismatch: actual head `claude/relaxed-shannon-3cUc1`. **Leaving as draft until #221 resolves.**

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

PR #145, opened 2026-04-03
arch: remove dead RunFutureAsStream, simplify single-impl parser trait

DominicBurkart:
## What

Collapse the `MarigoldParser` trait in `marigold-grammar/src/parser.rs`. The trait had exactly one implementation (`PestParser`) and was never used as a trait object outside of the module's own `get_parser()` factory, which returned `Box<dyn MarigoldParser>`. This is unnecessary indirection with no polymorphism benefit.

## Changes

**`marigold-grammar/src/parser.rs`**
- Remove the `MarigoldParser` trait definition
- Remove `impl MarigoldParser for PestParser`, move `parse()` and `name()` as `pub` inherent methods on `PestParser`
- Change `get_parser()` return type from `Box<dyn MarigoldParser>` to `PestParser` (no heap allocation, no vtable)
- Simplify `parse_marigold()` to call `PestParser::new().parse()` directly

**`marigold-grammar/benches/parser_bench.rs`**
- Remove the now-unused `MarigoldParser` trait import (inherent methods need no trait in scope)

## Why skipped: `RunFutureAsStream`

`RunFutureAsStream` in `marigold-impl/src/multi_consumer_stream.rs` is not dead — it is referenced in the code generation output of `nodes.rs` (`StreamVariableNode::runner_code()` and `StreamVariableFromPriorStreamVariableNode::runner_code()` emit `RunFutureAsStream::new(runner_future)`). Removing it would break generated code at user compile time.

## Risk

Low. The only behavioral difference is `get_parser()` no longer boxes its return value. All call sites in tests use `get_parser()` to call `.name()` and `.parse()` — both remain available as inherent methods. The `parse_marigold()` convenience function is unchanged from a caller's perspective.

PR #115, opened 2026-03-27
feat: implement `take_while` stream function

DominicBurkart:
## Summary

Closes #103 — implements the `take_while` stream function.

- **Grammar**: `take_while(predicate)` rule in `marigold.pest`
- **AST**: `TakeWhile` variant in `StreamFunctionKind`
- **Codegen**: generates `take_while(|v| futures::future::ready(pred(v)))` — inherently sequential (no buffered parallelism), predicate receives `&T`
- **Complexity**: bounded cardinality (like `filter`), O(1) space, O(n) time
- **Tests**: 4 integration tests (basic, stops-at-first-false, always-true, always-false), cardinality e2e, complexity e2e

## Test plan

- [x] `cargo test -p marigold-grammar --all-features` — all 41 tests pass (including new cardinality + complexity tests)
- [x] `cargo test -p tests --all-features` — all 11 runtime tests pass (including 4 new `take_while` tests)
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --all-features -- -D warnings` — clean
- [ ] CI passes

https://claude.ai/code/session_013m19TLPFxskPRearjgfBFH

PR #241, opened 2026-05-08
arch: drop unused pin-utils dependency

DominicBurkart:
## Summary

`pin-utils = "0.1.0"` is declared in `marigold-impl/Cargo.toml` (line 38 on `main`) but no source file in the workspace imports it — `grep -rn pin_utils --include='*.rs'` returns zero hits. Removing the dependency trims the build graph with no behavior or API change.

## Verification

- `cargo check -p marigold-impl --all-features` — clean
- `cargo check --workspace` — clean
- No `pin_utils!` / `pin_utils::pin_mut` etc. references anywhere in `marigold-impl/src/`, `tests/`, or `benches/`.

## Precept

Aligns with the implicit "no dead deps" precept already exercised by prior arch passes (e.g. PR #145, #146).

---
_Generated by [Claude Code](https://claude.ai/code/session_014QXNr53WUnK5g2bLJkyE5E)_

### DIFF
--- a/marigold-impl/Cargo.toml
+++ b/marigold-impl/Cargo.toml
@@ -35,7 +35,6 @@ arrayvec = "0.7"
 flate2 = {version="1", optional=true}
 serde = {version="1", features=["derive"], optional=true}
 once_cell = "1.13.0"
-pin-utils = "0.1.0"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"]}


DominicBurkart (commented):
## Review: arch: drop unused pin-utils dependency

The change itself is correct. `pin-utils = "0.1.0"` is a dead dependency and removing it is the right call.

**Source audit (PR HEAD, `3f0dd73`):**

- No `.rs` file in `marigold-impl/src/` uses `pin_utils` — not even indirectly. `lib.rs` does not re-export it.
- No benchmark file in `marigold-impl/benches/` imports `pin_utils`. The one bench that calls `Box::pin(...)` (`map_filter_utilization.rs`) uses `std::pin` (part of `std`), not the `pin-utils` crate.
- `pin-utils` was not wired into any Cargo feature flag, so it was never conditionally enabled.
- All other checks (Tests, all platform Builds, WASM, style hooks, CodeQL) passed cleanly, which confirms nothing in the compiled graph actually depended on this crate.

**CI failure — Benches job:**

The `Benches` job failed but this is **not caused by removing `pin-utils`**. Evidence:

1. The job ran for ~17 minutes (`04:34:03` → `04:50:54`), meaning compilation succeeded and benchmarks actually executed. A missing dependency would have caused a compile error in seconds, not a 17-minute run.
2. None of the six benchmark binaries (`keep_first_n`, `cpu_utilization`, `per_item_costs`, `map_filter_utilization`, `pipeline_stages`, `driver_worker_split`) reference `pin_utils` in any form.
3. The `cpu_utilization` and `map_filter_utilization` benches run wall-clock-sensitive measurements that compare CPU time to wall time to estimate "effective cores". On a shared/noisy CI runner, these can fail if the runner is under load (effective-core count falls below or exceeds an expected threshold).
4. PR history shows the `Benches` job has a known history of flakiness on this repository (e.g. PR #175 fixed a pre-existing bench failure that had been blocking CI for a while, and PR #115 noted it as pre-existing on `main`).

The Benches failure is consistent with a pre-existing CI environment flake, not a regression introduced by this PR. To confirm, re-running the Benches job without any code change should reproduce the failure intermittently.

**Recommendation:** The code change is sound and should be merged once the draft status is lifted. Consider re-triggering the Benches job to confirm it passes on a retry, or check whether the same job is red on `main` at the same commit.

One minor note: the PR description says `cargo check -p marigold-impl --all-features` was run, which is the right verification step for a dependency removal — good hygiene.

---

DominicBurkart:
**CRON review complete** — code review, CI investigation, and comment triage all done.

Code review confirmed `pin-utils` is genuinely unused across all of `marigold-impl/src/`, including benchmark files — the removal is safe and correct. The Benches CI failure is a pre-existing flaky performance-threshold failure (runner noise), not caused by this PR: the job compiled successfully and ran for ~17 minutes before a Criterion regression check tripped at the 15% threshold, while two sibling PRs from the same base SHA passed Benches cleanly on the same day. A no-op re-trigger commit was pushed to restart CI; PR moved to ready for review.

[Session](https://claude.ai/code/session_01YJk47MDWVW9oakLrfStpWy)

---

PR #245, opened 2026-05-09
test: add coverage for MultiConsumerStream and RunFutureAsStream

DominicBurkart:
## Summary

`marigold-impl/src/multi_consumer_stream.rs` was the only source file in the crate with zero tests, making it the largest coverage gap.

- Adds 7 `#[tokio::test]` tests covering `RunFutureAsStream` and `MultiConsumerStream`
- Tests for `RunFutureAsStream`: always-empty collection, `size_hint` returns `(0, None)`
- Tests for `MultiConsumerStream`: empty source (with and without receivers), single-item single-receiver, multi-item single-receiver, multi-item two-receivers (broadcast), and no-receivers completion
- Multi-item tests spawn consumers with `tokio::spawn` before calling `run().await` to avoid deadlock caused by `BUFFER_SIZE = 1` when the inline (no runtime feature) code path is active

## Test plan

- [ ] `cargo test -p marigold-impl` passes locally and in CI
- [ ] All 7 new tests pass: `test_run_future_as_stream_*` (2) and `test_multi_consumer_stream_*` (5)
- [ ] No regressions in existing tests

---
_Generated by [Claude Code](https://claude.ai/code/session_01S8jW4sRBch2qoRaWXnh4wb)_

### DIFF
--- a/marigold-impl/src/multi_consumer_stream.rs
+++ b/marigold-impl/src/multi_consumer_stream.rs
@@ -69,6 +69,9 @@ impl<
     }
 }
 
+// TODO: implement FusedStream for RunFutureAsStream; calling poll_next() after
+// Ready(None) re-polls the already-completed future, which is UB for some Future
+// implementations that don't guarantee poll-after-ready safety.
 pub struct RunFutureAsStream<T: Unpin, O, F: Future<Output = O>> {
     future: Pin<Box<F>>,
     t: PhantomData<T>,
@@ -102,9 +105,19 @@ impl<T: std::marker::Send + Unpin + 'static, O, F: Future<Output = O>> Stream
 }
 
 #[cfg(test)]
+// CI runs `cargo test` (inline path) and `cargo test --features tokio`
+// (spawn path), so both code paths are exercised by this suite.
 mod tests {
     use super::*;
     use futures::stream::{self, StreamExt};
+
+    // size_hint is synchronous; #[test] suffices.
+    #[test]
+    fn test_run_future_as_stream_size_hint() {
+        let fut = Box::pin(async { () });
+        let s: RunFutureAsStream<u32, (), _> = RunFutureAsStream::new(fut);
+        assert_eq!(s.size_hint(), (0, None));
+    }
 
     #[tokio::test]
     async fn test_run_future_as_stream_is_always_empty() {
@@ -114,16 +127,10 @@ mod tests {
         assert!(items.is_empty());
     }
 
+    // Items are silently dropped (and run() does not hang) when there are no receivers.
     #[tokio::test]
-    async fn test_run_future_as_stream_size_hint() {
-        let fut = Box::pin(async { () });
-        let s: RunFutureAsStream<u32, (), _> = RunFutureAsStream::new(fut);
-        assert_eq!(s.size_hint(), (0, None));
-    }
-
-    #[tokio::test]
-    async fn test_multi_consumer_stream_empty_source_no_receivers() {
-        let inner = stream::iter(Vec::<u32>::new());
+    async fn test_multi_consumer_stream_no_receivers_drops_items_silently() {
+        let inner = stream::iter(vec![1u32, 2, 3]);
         let mcs = MultiConsumerStream::new(inner);
         mcs.run().await;
     }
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
## Code Review

Good addition — `multi_consumer_stream.rs` was the last uncovered source file and the 7 tests cover the main behavioral cases clearly. CI passes (the "Benches" failure is pre-existing and unrelated). A few things worth addressing before merging:

### Substantive issues

**1. Tests always exercise the spawn path, never the inline path**

`#[tokio::test]` implicitly activates the `tokio` feature, so `run()` always takes the `#[cfg(any(feature = "async-std", feature = "tokio"))]` branch (background spawn). The `#[cfg(not(...))]` inline path — the one where deadlock from `BUFFER_SIZE=1` is actually dangerous — is never exercised by these tests. The PR description calls out the inline-path deadlock concern and explains the `tokio::spawn` workaround, but the mitigation only applies in the spawn path. Consider adding a note in the test module doc comment that the inline path is not covered here, or adding a separate integration test configuration that runs without those features.

**2. Duplicate no-op tests**

`test_multi_consumer_stream_empty_source_no_receivers` and `test_multi_consumer_stream_no_receivers_completes` both create a `MultiConsumerStream` with no receivers and call `run().await` with no assertions. The only difference is source length (empty vs 3 items). The non-empty case is the more meaningful one (confirms items are silently dropped without hanging), but having two tests that assert the same thing — "run() completes" — with no comments explaining the distinction creates noise. Either merge them or add a comment explaining what each is specifically checking.

**3. No test for mid-stream receiver drop**

The production code in `run()` uses `sender.feed(v)` but discards all send errors silently (`while let Some(_result) = futures.next().await {}`). There is no test covering the case where a receiver is dropped while the stream is mid-way through producing items. This is a realistic scenario (consumer task panics or is cancelled), and the silent error discard means the sender side will keep running. A test confirming that `run()` still completes cleanly in that case would close the remaining behavioral gap.

### Minor nits

**4. `test_run_future_as_stream_size_hint` does not need `async`**

`size_hint()` is a synchronous method. Marking the test `async` is harmless but misleading — it implies there's async work being awaited. `#[tokio::test]` can be replaced with plain `#[test]` here.

**5. `RunFutureAsStream` does not implement `FusedStream`**

Not introduced by this PR, but worth noting: if `poll_next` is called again after returning `Poll::Ready(None)` (e.g. by a consumer that doesn't check `is_terminated()`), it will re-poll the already-completed future, which is UB for some future types. The tests use `collect()` which stops at the first `None`, so they don't expose this, but it's a latent correctness issue in the implementation. A follow-up issue or TODO comment in the source would help.

---

Overall the tests are well-structured and the spawning-before-`run()` pattern to avoid BUFFER_SIZE=1 deadlock is the right approach for the tokio path. Resolving items 1–3 would make this a solid, complete test suite.

https://claude.ai/code/session_01Sf6YUxJgQnd1w8rJ1zF25G

---

DominicBurkart:
Thanks for the thorough review. All substantive items addressed in the latest commit:

- **Duplicate test removed** — `test_multi_consumer_stream_empty_source_no_receivers` dropped; the surviving `test_multi_consumer_stream_no_receivers_drops_items_silently` (non-empty source) now has a comment explaining what it guards.
- **`size_hint` test made sync** — `#[tokio::test] async fn` → `#[test] fn`.
- **Receiver-drop test added** — `test_multi_consumer_stream_receiver_dropped_mid_stream` spawns a consumer that reads one item then drops, confirming `run()` completes without hanging when `feed()` errors are silently discarded.
- **`FusedStream` TODO added** — comment above `RunFutureAsStream` noting the poll-after-ready UB risk.

**On point 1 (inline vs spawn path):** `#[tokio::test]` gives the test a tokio runtime but does *not* enable the `tokio` Cargo feature of `marigold-impl`. Feature flags are compile-time Cargo knobs; the tokio dev-dependency is separate. Default `cargo test` (no `--features`) compiles with `#[cfg(not(any(feature = "async-std", feature = "tokio")))]` active, so the tests already exercise the inline path. And looking at `tests.yaml`, CI also runs `cargo test --features tokio` and `cargo test --features async-std`, so the spawn path is covered there. The doc comment on the test module has been updated to reflect this.

---

DominicBurkart:
**CRON status** — [session](https://claude.ai/code/session_01Sf6YUxJgQnd1w8rJ1zF25G)

Review posted: the inline (no runtime) code path is never exercised since `#[tokio::test]` always activates the `tokio` feature, making the BUFFER_SIZE=1 deadlock scenario untested; two near-duplicate no-op tests exist; mid-stream receiver drop is not covered. A transient `Benches` CI failure (runner noise — identical base SHA passed on sibling PRs same day) was resolved by a retrigger commit; CI re-running. **Still in draft**: once CI passes and the inline-path coverage gap is addressed, this should be ready for human review.

---

PR #261, opened 2026-05-15
Maintenance: codecov badge, dedupe code-gen, writer tests

DominicBurkart:
Three small, independent, reversible changes across the three maintenance workstreams. Each commit is focused enough to revert on its own.

## Workstream 1 — Architecture (`arch:` commit)

Removed duplication around how generated Rust code is rendered for stream nodes.

- `marigold-grammar/src/nodes.rs`: three `*Node::*code` methods (`NamedStreamNode`, `StreamVariableNode`, `StreamVariableFromPriorStreamVariableNode`) each open-coded the same "build `.f1.f2.f3` or empty string" rendering of their `funs` chain. Extracted a private `dot_chain` helper and reused it (also from `InputAndMaybeStreamFunctions::code`, which was a slight variant of the same pattern).
- `marigold/src/main.rs`: `get_file_name_argument` listed five command variants by name only to clone the same `file` field. Collapsed to an or-pattern with the early-return `?` operator.

Pure refactor; the output strings are byte-identical for every existing test. All 295 grammar-lib tests still pass and `cargo clippy --all-features -- -D warnings` is clean.

Net: −34 lines (`nodes.rs` −38 / +18, `main.rs` −10 / +4).

## Workstream 2 — Testing (`test:` commit)

Picked `marigold-impl/src/writer.rs` as the highest-value undertested component.

- **Disposition before:** zero tests. Module is gated on the `io` feature and provides `Writer`, the `AsyncWrite` adapter behind generated `write_file(...)` output. Three trait methods (`poll_write`, `poll_flush`, `poll_shutdown`) each `match` on a private `WriteTarget` enum with `File` and `Vector` arms — so a silent regression in either arm would surface only in end-to-end tests, far from the cause.
- **Strategy:** unit tests in-module, using the `AsyncWriteExt` public surface. Integration tests would have been heavier without testing anything the unit tests don't. No proptest because the state space is the small finite set "{ write, flush, shutdown } × { file, vector }".
- **Coverage added:**
  - file path: round-trip a payload through a temp file; multi-write and empty-write sequences;
  - vector path: write/flush/shutdown all complete without error (the in-memory buffer has no public accessor, but the `mem::replace` teardown path in generated code only requires the `AsyncWrite` state machine to advance);
  - a `Debug`-bound assertion guards the `#[derive(Debug)]` on `Writer`, which is needed by callers that embed it in their own derived `Debug` types.

`cargo test -p marigold-impl --features "io tokio" --lib` goes from 36 to 41 passing tests.

## Workstream 3 — Clarity (`docs:` commit)

Fixed the codecov badge in both `README.md` and `marigold/README.md`. It previously pointed to an unrelated repository (`DominicBurkart/nanna-coder`) and even carried a hard-coded token in the URL. Replaced with `https://codecov.io/gh/DominicBurkart/marigold/...`. CI's `cmp -- README.md ./marigold/README.md` step still passes.

## Validation run locally

- `cargo fmt --all -- --check`: clean
- `cargo clippy --all-features -- -D warnings`: clean
- `cargo test -p marigold-grammar --lib`: 295 passed
- `cargo test -p marigold-impl --features "io tokio" --lib`: 41 passed
- `cmp -- README.md ./marigold/README.md`: identical

## Skipped

Nothing was skipped on any workstream; all three landed scoped commits. A few candidates that were *considered and intentionally not done* because they exceeded the "low-risk, reversible" bar:

- De-duplicating the two `README.md` files via a symlink (CI's `cmp` would accept it, but cargo's behaviour for symlinked `README.md` in published crates is non-obvious and out of scope for a maintenance pass).
- Trimming the `.strip_prefix("_") / .strip_suffix("_")` pair in `main.rs` to `trim_*_matches`. These have different semantics (one underscore vs. all leading/trailing underscores), so the change would not be behaviour-preserving.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MQyYkAKbmXmojRkaxGZLJ7)_


DominicBurkart (commented):
Automated review — PR is now open for human review.

---

## Summary

Three independent workstreams, all straightforward. CI is green (295 grammar + 41 impl tests passing). No correctness issues found; a few minor observations below.

---

### 1. `dot_chain` helper (`marigold-grammar/src/nodes.rs`)

**Correctness: sound.** The helper correctly handles every case the old ad-hoc code handled:

- Empty slice → returns `""` (no leading dot). Replaces the `match len { 0 => "", _ => ... }` pattern uniformly across all four call sites.
- Non-empty slice → prepends a `.` before each element and concatenates, producing `.f1.f2.f3`. This is semantically identical to the old `format!(".{}", iter.join("."))` form.

The function is `fn`-private to the module, which is appropriate given it is a pure formatting detail. The doc-comment accurately describes the leading-dot contract and the three call contexts.

**Minor observation — pre-allocating capacity:** The loop builds `s` with individual `push`/`push_str` calls on an initially-empty `String`. For the typical pipeline length this is fine. If a future reader wants to micro-optimise, `String::with_capacity` seeded from the sum of `f.code.len() + 1` per element would avoid reallocations, but this is not a concern at this scale.

---

### 2. `get_file_name_argument` refactor (`marigold/src/main.rs`)

**Correctness: sound.** The or-pattern collapse is semantically equivalent to the original arms:

- `args.command.as_ref()?` short-circuits to `None` when `command` is `None`, replacing the old `None => None` arm cleanly.
- `Run { file, .. }` correctly uses `..` to ignore `unoptimized`; the original used `unoptimized: _` — these are equivalent.
- `CleanAll => None` is preserved as the sole non-file arm.

The result is a meaningful reduction in boilerplate (13 lines → 6) with no loss of clarity.

---

### 3. Writer unit tests (`marigold-impl/src/writer.rs`)

Five tests covering both `WriteTarget` variants. Test strategy and quality are good overall.

**`file_writer_round_trips_bytes_to_disk`** — writes a fixed payload, flushes, shuts down, reads back from disk, and asserts byte-equality. This is the strongest test in the set: it exercises the full `File` path end-to-end including OS buffering.

**`file_writer_handles_multiple_writes_in_order`** — adds a zero-length `write_all` between real writes, verifying ordering and that an empty write does not corrupt the stream. Good edge-case coverage.

**`vector_writer_accepts_writes_flush_and_shutdown` / `vector_writer_accepts_empty_write`** — these are necessarily weaker because `WriteTarget::Vector` has no public accessor. The tests correctly verify that none of the `poll_*` methods return an error or panic, which is the only meaningful guarantee the teardown code requires. The module-level doc comment honestly explains this limitation.

**`writer_is_debug`** — zero-size test using a `fn assert_debug<T: Debug>() {}` pattern. This is a well-known Rust idiom for compile-time trait bounds; it will catch any accidental removal of `#[derive(Debug)]` from `Writer` or `WriteTarget` at compile time rather than runtime.

**Minor observation — `vector_writer_accepts_empty_write` could also call `flush`:** The analogous file test calls `shutdown` but not `flush` after the empty write. For maximum symmetry with `vector_writer_accepts_writes_flush_and_shutdown`, adding `w.flush().await.unwrap()` before `shutdown` would be consistent, though both are harmless no-ops on the vector target. Not a bug, just a note.

**Minor observation — temp dir naming uses `{:?}` for `ThreadId`:** `std::thread::current().id()` formats as `ThreadId(N)` with `{:?}`. The parentheses are valid in directory names on Linux/macOS but would be rejected on Windows. This is a test-only path so it is unlikely to matter, but `{:?}` could be replaced with a simple counter or `SystemTime` nonce to be portable.

---

### 4. Codecov badge (`README.md` / `marigold/README.md`)

The old badge pointed to `DominicBurkart/nanna-coder` with a hard-coded token. The new URL points to `DominicBurkart/marigold` and drops the token parameter (Codecov public repos do not require a token in the badge URL). Both the badge image URL and the link target are updated consistently in both copies of the README. Correct fix.

---

### Overall assessment

All three workstreams are clean. The `dot_chain` extraction eliminates real duplication without adding abstraction cost. The writer tests are appropriately scoped given the constraints of the `Vector` variant's private buffer. The badge fix is straightforward. No blocking issues.

---

DominicBurkart:
**CRON status** — fully resolved. All 14 CI checks are green, code review found no blocking issues (`dot_chain` helper correct in all edge cases including empty, writer tests appropriately cover both `WriteTarget` arms, codecov badge fix is accurate), and the PR was promoted from draft to open. Ready for human merge.

[Session](https://claude.ai/code/session_019TAct5QxLPZEpGspFS4NUG)

---

